PR #81, opened 2023-11-20
Add specific dispatch for `>=` constraints.

pedromxavier:
This PR is a first move towards avoiding bridges, as part of an initiative to better interact with JuMP's attribute system.


src/compiler/constraints.jl
```diff
@@ -160,6 +160,69 @@ function constraint(
...
+
+by adding a slack variable ``z``.
+"""
+function constraint(
+    model::Virtual.Model{T},
+    f::SAF{T},
+    s::GT{T},
+    arch::AbstractArchitecture,
+) where {T}
+    # Scalar Affine Inequality Constraint: g(x) = a'x - b ≥ 0
```
bernalde, line 194: Usually the notation always covers the <= case first. Why this choice?

src/compiler/parse.jl
```diff
@@ -152,3 +167,18 @@ function parse!(
...
+    s::GT{T},
+    arch::AbstractArchitecture,
+) where {T}
+    parse!(model, g, f, arch)
+
+    g[nothing] -= s.lower
+
+    return nothing
+end
+
```
bernalde, line 184: There are several warnings from codecov where lines are not covered by tests. Should we work on these before merging?

test/integration/examples/continuous/continuous_2.jl
```diff
@@ -0,0 +1,111 @@
...
+        # Solutions
+        x̂ = value.(x)
+        ŷ = objective_value(model)
+
+        @test x̂ ≈ x̄
+        @test ŷ ≈ ȳ
+    end
+
+    return nothing
+end
```
bernalde, line 111: It is a good test, but I like that tests also highlight examples from the literature so that we can always reference them. Can we use some of the ones that I showed in recent documents?

bernalde (commented):
Just leaving behind some comments, not to block the merge. Feel free to address them

src/compiler/constraints.jl
pedromxavier: The idea is that we have internal functions who will be shared by both cases. We want to avoid using bridges for this and handle them manually instead.

src/compiler/parse.jl
pedromxavier: Most are due to new attributes, it will be straightforward to cover them.

test/integration/examples/continuous/continuous_2.jl
pedromxavier: Of course!

PR #108, opened 2026-05-14
Implement linear constraint penalties

### LINKED ISSUE #90: Implement linear penalization method
Linear Ising penalty functions can be used instead of quadratic ones, as shown here https://doi.org/10.1103/PhysRevResearch.6.043241 and here https://doi.org/10.1088/1367-2630/ad7e4a
Their implementation in Python is available here https://collections.durham.ac.uk/files/r2fq977t82m
We should implement it as an option in toQUBO.

bernalde:
## Summary

- Add `QuadraticPenalty` and `LinearPenalty` constraint penalty method objects.
- Add model-level `DefaultConstraintEncodingMethod` and per-constraint `ConstraintEncodingMethod` attributes.
- Compile equality constraints with either the existing squared residual penalty or the new signed linear residual penalty; inequalities continue to use the existing quadratic slack formulation.
- Document the new settings and add unit/integration coverage for MOI and JuMP usage.

## Tests run

- `julia +1.12 --project=test -e 'push!(LOAD_PATH, pwd()); using Test; import MathOptInterface as MOI; using JuMP; const MOIU = MOI.Utilities; const VI = MOI.VariableIndex; const CI{F,S} = MOI.ConstraintIndex{F,S}; using QUBODrivers; using LinearAlgebra; using TOML; using ToQUBO; using ToQUBO: Attributes, Encoding; import QUBOTools; import PseudoBooleanOptimization as PBO; function MOIU.map_indices(::Function, arch::QUBOTools.AbstractArchitecture) return arch end; function MOIU.map_indices(::Function, quad::PBO.QuadratizationMethod) return quad end; function QUBOTools.backend(model::JuMP.Model) return QUBOTools.backend(JuMP.unsafe_backend(model)) end; include("test/unit/compiler/constraints.jl"); include("test/unit/attributes/attributes.jl"); @testset "targeted linear penalty" begin test_compiler_constraints_linear_penalty(); test_compiler_attributes(); test_constraint_attributes(); end'`
  - Passed: 92/92.
- `julia +1.12 --project=. -e 'using Pkg; Pkg.test()'`
  - Passed: 726/726.
- `julia +1.12 --project=docs docs/make.jl`
  - Passed. Documenter reported existing warnings for docstrings not included in the manual and skipped deployment because the command was run outside GitHub Actions.
- `git diff --check`
  - Passed.

## Notes

- No separate lint, type-check, or formatting command is documented in the repository. The repo has `.JuliaFormatter.toml`, but JuliaFormatter is not listed as a project/test/docs dependency.
- The default `julia` executable is 1.11.5 locally; `julia --project=. -e 'using Pkg; Pkg.test()'` failed before loading ToQUBO because the manifest was resolved with Julia 1.12 and `OpenSSL_jll` could not be located. The suite was rerun successfully with `julia +1.12`.

Closes #90

### DIFF
--- a/src/compiler/constraints.jl
+++ b/src/compiler/constraints.jl
@@ -26,6 +26,18 @@ function constraints!(
     return nothing
 end
 
+function _is_quadratic_penalty(model::Virtual.Model, ci::CI)::Bool
+    return Attributes.constraint_encoding_method(model, ci) isa Attributes.QuadraticPenalty
+end
+
+function _equality_penalty(model::Virtual.Model, ci::CI, g::PBO.PBF)
+    if _is_quadratic_penalty(model, ci)
+        return g^2
+    else
+        return g
+    end
+end
+
 @doc raw"""
     constraint(
         ::Virtual.Model{T},
@@ -80,7 +92,7 @@ into
 """
 function constraint(
     model::Virtual.Model{T},
-    ::CI,
+    ci::CI,
     f::SAF{T},
     s::EQ{T},
     arch::AbstractArchitecture,
@@ -105,7 +117,7 @@ function constraint(
         @warn "Infeasible constraint detected"
     end
 
-    return g^2
+    return _equality_penalty(model, ci, g)
 end
 
 @doc raw"""
@@ -283,7 +295,7 @@ into
 """
 function constraint(
     model::Virtual.Model{T},
-    ::CI,
+    ci::CI,
     f::SQF{T},
     s::EQ{T},
     arch::AbstractArchitecture,
@@ -311,10 +323,12 @@ function constraint(
         """
     end
 
-    # Tell the compiler that quadratization is necessary
-    MOI.set(model, Attributes.Quadratize(), true)
+    if _is_quadratic_penalty(model, ci)
+        # Tell the compiler that quadratization is necessary
+        MOI.set(model, Attributes.Quadratize(), true)
+    end
 
-    return g^2
+    return _equality_penalty(model, ci, g)
 end
 
 
--- a/docs/src/booklet/4-encoding.md
+++ b/docs/src/booklet/4-encoding.md
@@ -93,4 +93,12 @@ A QUBO model is unconstrained. So when `ToQUBO` is reformulating a problem, it n
 As constraints are introduced into the objective function, we need to make sure that they won't be violated.
 In order to do that, `ToQUBO` multiplies the encoded constraint by a large penalty ``\rho``, so that any violation would result in a sub-optimal solution to the problem.
 
+Equality constraints use a squared residual penalty by default. The compiler can
+also encode an equality constraint with a signed linear residual through
+[`ToQUBO.Attributes.LinearPenalty`](@ref). This is a heuristic method: the sign
+and magnitude of ``\rho`` affect whether the residual discourages the intended
+violations, so `ToQUBO` requires an explicit
+[`ToQUBO.Attributes.ConstraintEncodingPenaltyHint`](@ref) instead of inferring
+``\rho`` automatically.
+
 Sometimes, the encoding process might introduce higher-order terms, demanding `ToQUBO` to reduce the offending polynomials back to a quadratic form.
--- a/docs/src/booklet/4-encoding.md
+++ b/docs/src/booklet/4-encoding.md
@@ -95,10 +95,24 @@ In order to do that, `ToQUBO` multiplies the encoded constraint by a large penal
 
 Equality constraints use a squared residual penalty by default. The compiler can
 also encode an equality constraint with a signed linear residual through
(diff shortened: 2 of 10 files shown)


src/compiler/constraints.jl
```diff
@@ -26,6 +26,18 @@ function constraints!(
...
 
+function _is_quadratic_penalty(model::Virtual.Model, ci::CI)::Bool
+    return Attributes.constraint_encoding_method(model, ci) isa Attributes.QuadraticPenalty
+end
+
+function _equality_penalty(model::Virtual.Model, ci::CI, g::PBO.PBF)
+    if _is_quadratic_penalty(model, ci)
+        return g^2
+    else
+        return g
```
bernalde, line 37: Blocking: Returning the signed residual here is only half of the linear-penalty change. `penalties!` still treats every constraint penalty as if it were a nonnegative squared residual and auto-generates `rho` when `ConstraintEncodingPenaltyHint` is absent. With `LinearPenalty()`, that silently chooses an arbitrary signed field and can prefer infeasible assignments; for example, minimizing `0` with `x1 + x2 == 1` compiles to `-1 + x1 + x2` with `rho = 1`, so `x = (0, 0)` is cheapest. Since the existing API auto-computes penalties when hints are absent, selecting this new method currently produces a broken QUBO by default. Please make the coefficient path method-aware: either require an explicit `ConstraintEncodingPenaltyHint` for `LinearPenalty()` with a clear compilation error, or implement a documented/tuned linear penalty search, and add an `optimize!`-level regression test for that behavior.

bernalde (commented):
Blocking review finding: the linear residual encoding currently reuses the existing squared-penalty auto-scaling path when no `ConstraintEncodingPenaltyHint` is supplied, which can make infeasible assignments cheaper by default. I attempted to submit this as REQUEST_CHANGES, but GitHub does not allow this account to request changes on its own pull request. See the inline comment for the required fix.

bernalde:
Pushed commit `5b87c39` (`Require hints for linear penalties`).

Main changes made:
- Added a method-aware guard in `src/compiler/penalties.jl` so equality constraints using `LinearPenalty()` cannot fall back to the squared-penalty auto-scaling heuristic.
- `LinearPenalty()` now requires an explicit `ConstraintEncodingPenaltyHint`; inequalities still keep the existing quadratic slack formulation and automatic penalty inference.
- Updated docstrings for `LinearPenalty()` and `ConstraintEncodingMethod()` to document the explicit hint requirement.
- Added an optimize-level regression test that throws without a hint and compiles once a hint is provided.

Tests run and results:
- Targeted linear penalty tests: `julia +1.12 --project=test -e ...` passed, 97/97.
- Full package suite: `julia +1.12 --project=. -e 'using Pkg; Pkg.test()'` passed, 731/731.
- Docs build: `julia +1.12 --project=docs docs/make.jl` passed. It still reports the existing unrelated Documenter warnings about docstrings not included in the manual and skipped deployment outside CI.
- `git diff --check` passed.

- None.

Remaining risks or follow-up items:
- The linear penalty method remains heuristic; users still need to choose an appropriate signed penalty hint for their problem instance.

src/compiler/constraints.jl
bernalde: Addressed in 5b87c39. `src/compiler/penalties.jl` now rejects `LinearPenalty()` on equality constraints unless `ConstraintEncodingPenaltyHint` is set, so it no longer reuses the squared-penalty auto-scaling heuristic. Added an optimize-level regression test in `test/unit/compiler/constraints.jl` and updated the relevant docstrings.

bernalde (commented):
Review summary:

1. Blocking issues: none.
2. Nonblocking issues: none.
3. Questions: none.
4. Tests run and outcomes: focused linear penalty compiler tests passed 9/9; focused JuMP interface tests passed 86/86; docs build passed with only existing unrelated Documenter warnings; full package suite passed 731/731; git diff --check passed. Current GitHub checks are passing for Ubuntu, Windows, documentation, documenter/deploy, and labels.
5. Merge as-is: yes. I attempted to submit APPROVE, but GitHub rejects approvals from the PR author account.

bernalde (commented):
Review summary:

1. Blocking issues: none.
2. Nonblocking issues: none.
3. Questions: none.
4. Tests run and outcomes: discovered documented commands from .github/copilot-instructions.md and CI; focused linear penalty compiler tests passed 9/9; focused JuMP interface tests passed 86/86; docs build passed with only existing unrelated Documenter warnings; full package suite passed 731/731; git diff --check passed. GitHub checks are passing for Ubuntu CI, Windows CI, documentation, documenter/deploy, and label.
5. Merge as-is: yes. I attempted to submit APPROVE, but GitHub rejects approvals from the PR author account.

PR #114, opened 2026-05-17
Fix constraint feasibility actions

### LINKED ISSUE #46: Constraint Feasibility Checks & Actions
When generating penalty functions for constraints like  $$a x + b \le 0$$ i.e. $$g(x) \le 0$$ one might go for $$\rho_k (g_k(x) + s)^{2} = 0, g(x) \in \text{PBF}$$ to  compose the objective function.

As it is expected from PBF's that go $$ g: \mathbb{B}^{n} \to [a, b] $$, we can compute $$ [l, u] $$ such that $$ [a, b] \subseteq [l, u] $$ which gives us checks on wheter the function is infeasible or even always feasible. Tests against [MOI's Knapsack](https://jump.dev/MathOptInterface.jl/stable/tutorials/example/) were successful in detecting this case.

The question is: what to do when such things happen?

My first suggestion is:
```julia
# -*- Bounds & Slack Variable -*-
l, u = PBO.bounds(g)

s = if u < zero(T) # Always feasible
    @warn "Always-feasible constraint detected"
    continue
elseif l > zero(T) # Infeasible
    @warn "Infeasible constraint detected"
    error()
else
    VM.expansion(VM.encode!(VM.Binary, model, nothing, zero(T), abs(l)))
end
```

Although, I also suggest for `continue` and `error()` to be triggered by flags such as `:ignore_feasible` and `:error_infeasible`. Setting MOI termination status seems appropriate in the latter.

bernalde:
## Summary

- Added compiler attributes to control feasibility actions for provably always-feasible and infeasible scalar constraints.
- Centralized scalar residual feasibility checks and made strict direct infeasibility stop compilation with `MOI.INFEASIBLE`.
- Kept infeasible indicator inner constraints gated by their activation instead of treating them as globally infeasible.

## Tests run

- `julia +1.12 --project=test --startup-file=no -e 'push!(LOAD_PATH, "."); using Test; using JuMP; import MathOptInterface as MOI; const MOIU = MOI.Utilities; const VI = MOI.VariableIndex; const CI{F,S} = MOI.ConstraintIndex{F,S}; using QUBODrivers; using LinearAlgebra; using TOML; using ToQUBO; using ToQUBO: Attributes, Encoding; import QUBOTools; import PseudoBooleanOptimization as PBO; function MOIU.map_indices(::Function, arch::QUBOTools.AbstractArchitecture); return arch; end; function MOIU.map_indices(::Function, quad::PBO.QuadratizationMethod); return quad; end; include("test/unit/compiler/constraints.jl"); include("test/unit/attributes/attributes.jl"); @testset "targeted issue 46" begin test_compiler_constraints(); test_compiler_attributes(); end'`
  - Result: passed, 135/135 assertions.
- `julia +1.12 --project=. --startup-file=no -e 'using Pkg; Pkg.test()'`
  - Result: passed, 885/885 tests.

## Notes

- I did not use the local default `julia --project=.` for validation because the default Julia 1.11 environment fails while precompiling MathOptInterface via PrecompileTools (`UndefVarError: StaticData not defined in Base`). Julia 1.12 validation completed successfully.

Closes #46

### DIFF
--- a/src/compiler/constraints.jl
+++ b/src/compiler/constraints.jl
@@ -60,6 +60,58 @@ function _combine_penalties(lhs, rhs)
     end
 end
 
+abstract type _ConstraintFeasibilityContext end
+
+struct _DirectConstraintContext <: _ConstraintFeasibilityContext end
+struct _IndicatorConstraintContext <: _ConstraintFeasibilityContext end
+
+const _DIRECT_CONSTRAINT = _DirectConstraintContext()
+const _INDICATOR_CONSTRAINT = _IndicatorConstraintContext()
+
+function _constraint_message(kind::AbstractString, f, op::AbstractString, rhs)
+    return """
+    $(kind) constraint detected:
+    $(f) $(op) $(rhs)
+    """
+end
+
+function _constraint_warning(model::Virtual.Model, message::AbstractString)
+    if Attributes.warnings(model)
+        @warn message
+    end
+
+    return nothing
+end
+
+function _infeasible_constraint_error!(model::Virtual.Model, message::AbstractString)
+    MOI.set(model, Attributes.CompilationStatus(), MOI.INFEASIBLE)
+    MOI.set(model, MOI.RawStatusString(), message)
+
+    compilation_error(message)
+
+    return nothing
+end
+
+function _handle_feasible_constraint(model::Virtual.Model, message::AbstractString)::Bool
+    _constraint_warning(model, message)
+
+    return Attributes.ignore_feasible_constraints(model)
+end
+
+function _handle_infeasible_constraint(
+    model::Virtual.Model,
+    context::_ConstraintFeasibilityContext,
+    message::AbstractString,
+)
+    _constraint_warning(model, message)
+
+    if context isa _DirectConstraintContext && Attributes.error_infeasible_constraints(model)
+        _infeasible_constraint_error!(model, message)
+    end
+
+    return nothing
+end
+
 @doc raw"""
     constraint(
         ::Virtual.Model{T},
@@ -130,6 +182,17 @@ function constraint(
     s::EQ{T},
     arch::AbstractArchitecture,
 ) where {T}
+    return constraint(model, ci, f, s, arch, _DIRECT_CONSTRAINT)
+end
+
+function constraint(
+    model::Virtual.Model{T},
+    ci::CI,
+    f::SAF{T},
+    s::EQ{T},
+    arch::AbstractArchitecture,
+    context::_ConstraintFeasibilityContext,
+) where {T}
     # Scalar Affine Equality Constraint: g(x) = a'x - b = 0
     g = _parse(model, f, s, arch)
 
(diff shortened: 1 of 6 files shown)


src/compiler/constraints.jl
```diff
@@ -60,6 +60,58 @@ function _combine_penalties(lhs, rhs)
...
+    if Attributes.warnings(model)
+        @warn message
+    end
+
+    return nothing
+end
+
+function _infeasible_constraint_error!(model::Virtual.Model, message::AbstractString)
+    MOI.set(model, Attributes.CompilationStatus(), MOI.INFEASIBLE)
+    MOI.set(model, MOI.RawStatusString(), message)
```
bernalde, line 88: Blocking: This sets the local raw status, but users with an inner optimizer will not see it. `MOI.get(::Virtual.Model, ::MOI.RawStatusString)` currently delegates to `model.optimizer` whenever the inner optimizer supports the attribute, so strict infeasible compilation stops before the inner optimizer runs and `RawStatusString()` can come back as `""`. I reproduced this with `ToQUBO.Optimizer{Float64}(ExactSampler.Optimizer)`: `CompilationStatus()` and `TerminationStatus()` are `MOI.INFEASIBLE`, but `RawStatusString()` is empty. This matters because the new strict infeasible mode promises to report the detected model infeasibility before solver handoff. Please make the raw-status getter prefer `model.moi_settings[:raw_status_string]` when compilation status is not `OPTIMIZE_NOT_CALLED` or `LOCALLY_SOLVED` (or whenever a local raw status has been explicitly set before delegating), and add a solver-backed regression test for this path.

bernalde (commented):
I found one blocking status-reporting regression in the strict infeasible path. I attempted to submit this as REQUEST_CHANGES, but GitHub rejects request-changes reviews from the PR author account. The package tests pass, but the new behavior does not preserve RawStatusString for normal solver-backed usage.

bernalde:
Pushed commit:
- `6c4e3d0` Preserve compile raw status for infeasible constraints

Main changes made:
- Addressed the unresolved blocking review thread about strict infeasible `RawStatusString` reporting.
- Updated `MOI.get(::Virtual.Model, ::MOI.RawStatusString)` to prefer locally recorded compile-failure status before delegating to an inner optimizer.
- Added solver-backed regression coverage for strict infeasible compilation with `ExactSampler.Optimizer`.

Tests run:
- `git diff --check`: passed.
- `julia +1.12 --project=test --startup-file=no -e 'push!(LOAD_PATH, "."); using Test; using JuMP; import MathOptInterface as MOI; const MOIU = MOI.Utilities; const VI = MOI.VariableIndex; const CI{F,S} = MOI.ConstraintIndex{F,S}; using QUBODrivers; using LinearAlgebra; using TOML; using ToQUBO; using ToQUBO: Attributes, Encoding; import QUBOTools; import PseudoBooleanOptimization as PBO; function MOIU.map_indices(::Function, arch::QUBOTools.AbstractArchitecture); return arch; end; function MOIU.map_indices(::Function, quad::PBO.QuadratizationMethod); return quad; end; include("test/unit/compiler/constraints.jl"); include("test/unit/attributes/attributes.jl"); @testset "targeted review fix" begin test_compiler_constraints(); test_compiler_attributes(); end'`: passed, 140/140.
- `julia +1.12 --project=. --startup-file=no -e 'using Pkg; Pkg.test()'`: passed, 890/890.

- None.

Remaining risks or follow-up:
- None identified locally; GitHub Actions will rerun on the pushed commit.

src/compiler/constraints.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/6c4e3d0. The raw-status getter now returns locally recorded compile-failure status before delegating to an inner optimizer, and `test_compiler_constraints_strict_infeasible_status` covers the `ExactSampler.Optimizer` path.

bernalde (commented):
Focused status-reporting review: no findings. The 6c4e3d0 change preserves local compile-failure RawStatusString before inner optimizer delegation, while successful solver-backed optimize still delegates RawStatusString to the inner optimizer. Targeted status tests and the full package suite passed. I attempted to submit this as APPROVE, but GitHub rejected approval from the PR author account.

Issue #132, opened 2026-06-05
Record reformulation metadata for original and auxiliary variables


bernalde:
## Motivation

The DS-MFG QUBO contains original flow variables plus auxiliary/slack variables introduced by reformulation. The Gurobi pool is naturally expressed in the original variables, while the sampled QUBO state includes auxiliary bits. Comparing the two required notebook-local knowledge of which bits were original and how to repair auxiliaries.

ToQUBO is the package that creates these encodings, so it should record enough reformulation metadata for downstream solvers and analysis tools to interpret samples.

## Proposed PR scope

Record structured reformulation metadata during compilation:

- original variable identities and ordering;
- generated auxiliary/slack variable identities and ordering;
- mapping from original variables to QUBO variables;
- scale and offset introduced by reformulation;
- penalty terms and constraint encodings, where feasible;
- enough metadata to project a QUBO state back to original variables;
- enough metadata to identify which auxiliary variables belong to which original constraint/encoding.

Expose helper functions such as:

- `original_variables(model_or_metadata)`;
- `auxiliary_variables(model_or_metadata)`;
- `project_original_state(model_or_metadata, qubo_state)`;
- `is_encoded_consistently(model_or_metadata, qubo_state)`, if feasible;
- a hook point for exact auxiliary repair when the encoding supports it.

## Acceptance criteria

- Metadata is attached to the QUBOTools model metadata in a JSON-compatible structure.
- Existing ToQUBO outputs remain backwards compatible for users who do not inspect metadata.
- Tests cover at least one constrained binary/integer model where auxiliary variables are created.
- Projection from full QUBO state to original variables is tested.
- Documentation explains what metadata is guaranteed and what remains encoding-specific.

## Package-boundary decision

This belongs in `ToQUBO.jl`, not `QiskitOpt.jl`, because it concerns how the original optimization problem was reformulated into QUBO variables. QiskitOpt should consume this metadata but should not infer it.

bernalde:
Implemented by PR #133 and released in v0.4.0. Closing this tracker; follow-up repair semantics can be split into a narrower issue if needed.

PR #133, opened 2026-06-06
Record reformulation metadata

bernalde:
## Summary

- Add JSON-compatible ToQUBO reformulation metadata with schema versioning, source/target variable order, expansion terms, auxiliary ownership, slack encodings, and compiler penalty terms.
- Attach reformulation metadata to the `QUBOTools.backend` model under `metadata["toqubo"]["reformulation"]`.
- Add helper APIs for reading original variables, auxiliary variables, and projecting a full QUBO state back onto original variables.
- Document the metadata guarantees and the encoding-specific limits around auxiliary consistency/repair.

## Tests

- `julia +release --project=. -e 'using Test; import MathOptInterface as MOI; using ToQUBO; using ToQUBO: Attributes; import QUBOTools; const VI = MOI.VariableIndex; include("test/unit/reformulation.jl"); test_reformulation_metadata()'`
- `julia +release --project=. -e 'import Pkg; Pkg.test()'`
- `julia +release --project=docs docs/make.jl`

Refs #132

### DIFF
--- a/docs/src/manual/3-results.md
+++ b/docs/src/manual/3-results.md
@@ -152,13 +152,16 @@ zeta = MOI.get(model, Attributes.SlackVariableEncodingFunction(), constraint_ref
 ### Reformulation Metadata
 
 ToQUBO records JSON-compatible reformulation metadata for downstream tools that
-need to interpret full QUBO states. The metadata is available directly from the
-optimizer and is also attached under `metadata["toqubo"]["reformulation"]` on
-the `QUBOTools` model returned by `QUBOTools.backend`.
+need to interpret full QUBO states. The metadata uses string-keyed dictionaries,
+arrays, primitive values, and finite numeric values. It is available directly
+from the optimizer and is also attached under
+`metadata["toqubo"]["reformulation"]` on the `QUBOTools` model returned by
+`QUBOTools.backend`.
 
 ```julia
-backend = unsafe_backend(model)
-metadata = ToQUBO.reformulation_metadata(backend)
+optimizer = unsafe_backend(model)
+metadata = ToQUBO.reformulation_metadata(optimizer)
+qubo_model = ToQUBO.QUBOTools.backend(optimizer)
 
 original = ToQUBO.original_variables(metadata)
 auxiliary = ToQUBO.auxiliary_variables(metadata)
--- a/docs/src/manual/3-results.md
+++ b/docs/src/manual/3-results.md
@@ -149,7 +149,36 @@ z = MOI.get(model, Attributes.SlackVariableTargetVariables(), constraint_ref)
 zeta = MOI.get(model, Attributes.SlackVariableEncodingFunction(), constraint_ref)
 ```
 
+### Reformulation Metadata
+
+ToQUBO records JSON-compatible reformulation metadata for downstream tools that
+need to interpret full QUBO states. The metadata is available directly from the
+optimizer and is also attached under `metadata["toqubo"]["reformulation"]` on
+the `QUBOTools` model returned by `QUBOTools.backend`.
+
+```julia
+backend = unsafe_backend(model)
+metadata = ToQUBO.reformulation_metadata(backend)
+
+original = ToQUBO.original_variables(metadata)
+auxiliary = ToQUBO.auxiliary_variables(metadata)
+
+qubo_state = [0, 1, 0, 1]
+source_state = ToQUBO.project_original_state(metadata, qubo_state)
+```
+
+The metadata guarantees the original source variable order, target QUBO
+variable order, source-to-target `expansion_variables` and `expansion_terms`,
+constraint/slack ownership for generated target variables, and penalty terms
+recorded by the compiler. This is enough to project a full QUBO state back to
+original variables. Exact auxiliary consistency checks and repair remain
+encoding-specific and are not guaranteed by this metadata contract.
+
 ```@docs
+ToQUBO.reformulation_metadata
+ToQUBO.original_variables
+ToQUBO.auxiliary_variables
+ToQUBO.project_original_state
 ToQUBO.Attributes.CompilationTime
 ToQUBO.Attributes.CompilationStatus
 ToQUBO.Attributes.SourceModel
--- a/src/reformulation.jl
+++ b/src/reformulation.jl
@@ -184,6 +184,8 @@ function _target_metadata(model::Virtual.Model{T}) where {T}
     variables = MOI.get(model.target_model, MOI.ListOfVariableIndices())
     objective = MOI.get(model.target_model, MOI.ObjectiveFunction{SQF{T}}())
 
+    # ToQUBO adds penalty and quadratization terms, but does not rescale the
+    # source objective during reformulation.
     return Dict{String,Any}(
         "num_variables" => length(variables),
         "variable_order" => [vi.value for vi in variables],
@@ -197,6 +199,7 @@ function _variable_encoding_entry(model::Virtual.Model, vi::VI)
     entry = Dict{String,Any}(
         "id" => vi.value,
         "name" => _variable_name(model.source_model, vi),
+        "encoded" => false,
         "target_variables" => Int[],
         "encoding" => nothing,
         "expansion_variables" => Int[],
@@ -208,6 +211,7 @@ function _variable_encoding_entry(model::Virtual.Model, vi::VI)
     if haskey(model.source, vi)
         v = model.source[vi]
 
+        entry["encoded"] = true
         entry["target_variables"] = [yi.value for yi in Virtual.target(v)]
         entry["encoding"] = _encoding_metadata(Virtual.encoding(v))
         entry["expansion_variables"] = _pbf_variable_ids(Virtual.expansion(v))
@@ -460,6 +464,10 @@ function _evaluate_serialized_terms(terms::AbstractVector, state)
     return value
 end
 
+function _serialized_entry_encoded(entry::AbstractDict)
+    return get(entry, "encoded", !isempty(entry["expansion_terms"]))
+end
+
 function project_original_state(model::Virtual.Model, qubo_state)
     values = Dict{VI,Any}()
 
@@ -483,6 +491,10 @@ function project_original_state(metadata::AbstractDict, qubo_state)
     values = Dict{Int,Any}()
 
     for entry in data["original_variables"]
+        if !_serialized_entry_encoded(entry)
+            error("Source variable $(entry["id"]) has not been encoded")
+        end
+
         values[entry["id"]] = _evaluate_serialized_terms(
             entry["expansion_terms"],
             qubo_state,
--- a/src/reformulation.jl
+++ b/src/reformulation.jl
@@ -1,0 +1,497 @@
+const _TOQUBO_METADATA_KEY = "toqubo"
+const _REFORMULATION_METADATA_KEY = "reformulation"
+const _REFORMULATION_SCHEMA_VERSION = 1
+
+@doc raw"""
+    reformulation_metadata(model_or_metadata)
+
+Return JSON-compatible ToQUBO reformulation metadata. For a live
+`ToQUBO.Optimizer`, the metadata is generated from the compiled virtual model.
+For a `QUBOTools.AbstractModel` or a metadata dictionary, the metadata is read
+from `metadata["toqubo"]["reformulation"]`.
+"""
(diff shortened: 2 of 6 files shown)


src/reformulation.jl
```diff
@@ -0,0 +1,497 @@
...
+function project_original_state(model::QUBOTools.AbstractModel, qubo_state)
+    return project_original_state(reformulation_metadata(model), qubo_state)
+end
+
+function project_original_state(metadata::AbstractDict, qubo_state)
+    data = reformulation_metadata(metadata)
+    values = Dict{Int,Any}()
+
+    for entry in data["original_variables"]
+        values[entry["id"]] = _evaluate_serialized_terms(
```
bernalde, line 498: Nonblocking: this serialized path silently returns 0 when a variable's `expansion_terms` is empty (an unencoded source variable), because `_evaluate_serialized_terms` over an empty list returns 0. The live `project_original_state(model::Virtual.Model, ...)` path errors in the same case (line 468: `error("Source variable ... has not been encoded")`). Same operation, divergent behavior. Suggest erroring or warning here too when `expansion_terms` is empty, or documenting that unencoded originals project to 0.

src/reformulation.jl
```diff
@@ -0,0 +1,497 @@
...
+
+function _target_metadata(model::Virtual.Model{T}) where {T}
+    variables = MOI.get(model.target_model, MOI.ListOfVariableIndices())
+    objective = MOI.get(model.target_model, MOI.ObjectiveFunction{SQF{T}}())
+
+    return Dict{String,Any}(
+        "num_variables" => length(variables),
+        "variable_order" => [vi.value for vi in variables],
+        "sense" => _sense_metadata(MOI.get(model.target_model, MOI.ObjectiveSense())),
+        "scale" => one(T),
```
bernalde, line 193: Nonblocking: `scale` is hardcoded to `one(T)`. Issue #132 asks to record the scale introduced by reformulation. Penalty reformulation does not rescale the objective so 1 is correct today, but it is asserted rather than derived. Add a comment explaining why it is always 1, or derive it from the target model so a future scaling encoding cannot silently record a wrong value.

docs/src/manual/3-results.md
```diff
@@ -149,7 +149,36 @@ z = MOI.get(model, Attributes.SlackVariableTargetVariables(), constraint_ref)
 zeta = MOI.get(model, Attributes.SlackVariableEncodingFunction(), constraint_ref)
 ```
 
+### Reformulation Metadata
+
+ToQUBO records JSON-compatible reformulation metadata for downstream tools that
```
bernalde, line 154: Nonblocking: the "JSON-compatible" guarantee is verified only structurally by `_json_compatible` in the test, with no actual serialize/parse round-trip (JSON is not a dependency). That proxy will not catch non-finite coefficients (Inf/NaN from large penalties are not valid JSON). Consider a real round-trip test or a note that the structural check is intentional.

docs/src/manual/3-results.md
```diff
@@ -149,7 +149,36 @@ z = MOI.get(model, Attributes.SlackVariableTargetVariables(), constraint_ref)
...
+need to interpret full QUBO states. The metadata is available directly from the
+optimizer and is also attached under `metadata["toqubo"]["reformulation"]` on
+the `QUBOTools` model returned by `QUBOTools.backend`.
+
+```julia
+backend = unsafe_backend(model)
```
bernalde: Question/minor: this names the JuMP `unsafe_backend(model)` result `backend`, but the prose just above discusses `QUBOTools.backend`, which is a different function returning a different object (the `QUBOTools.Model`). Consider renaming the variable or qualifying the call to avoid conflating the two.

bernalde (commented):
Review of #133 (record reformulation metadata).

Note: I am the author of this PR, so GitHub does not permit me to submit APPROVE or REQUEST_CHANGES on it. This review is therefore submitted as COMMENT. There are no blocking issues; the formal approval must come from another maintainer.

Summary of what the PR does: adds `src/reformulation.jl` building a JSON-compatible reformulation metadata dict from the compiled `Virtual.Model`, attaches it to the `QUBOTools` model under `metadata["toqubo"]["reformulation"]` via `QUBOTools.backend`, and adds reader/helper APIs (`reformulation_metadata`, `original_variables`, `auxiliary_variables`, `project_original_state`) that work on a live optimizer, a `QUBOTools.AbstractModel`, or a serialized metadata dict.

Verdict against the motivating issue: issue #132 lists `is_encoded_consistently` and a hook for exact auxiliary repair, both hedged "if feasible." This PR does not implement those and documents that auxiliary consistency/repair remain encoding-specific. The PR correctly links with `Refs #132` (not `Closes`), so it will not wrongly auto-close the issue on merge. That is the right call for a partial-by-design implementation.

BLOCKING ISSUES

None.

NONBLOCKING ISSUES

1. Divergent error handling between the live and serialized `project_original_state` paths (src/reformulation.jl). The live path errors when a source variable was not encoded; the serialized path silently returns 0 for the same situation (empty `expansion_terms` evaluates to 0). Same logical operation, two different behaviors. Fix: make the serialized path also error (or at least warn) when a variable's `expansion_terms` is empty, or document that unencoded originals project to 0.

2. `"scale" => one(T)` is hardcoded in `_target_metadata` (src/reformulation.jl:191). The issue asks to record the scale introduced by reformulation. ToQUBO's penalty reformulation does not rescale the objective, so 1 is correct in practice, but it is asserted rather than derived/validated. Add a short comment explaining why scale is always 1, or derive it from the target model so a future scaling encoding does not silently record a wrong value.

3. The "JSON-compatible" guarantee is verified only structurally by `_json_compatible` in the test; there is no real serialize/parse round-trip (JSON is not a project or test dependency). The structural check is a reasonable pragmatic proxy, but it will not catch non-finite coefficients (Inf/NaN from large penalties are not representable in JSON). Optional: add JSON as a test dep and assert an actual round-trip, or note the structural check is intentional.

4. Minor type inconsistency: `_evaluate_serialized_terms` initializes `value = 0` (Int) so an empty term list returns an Int while a non-empty one returns the coefficient type (Float). Harmless for current callers, but inconsistent.

QUESTIONS

1. Confirm #132 is intended to stay open after merge for the deferred `is_encoded_consistently` / auxiliary-repair work. The `Refs #132` link is correct for that intent; just confirming it is deliberate rather than an oversight.

2. In the doc example the JuMP `unsafe_backend(model)` result is named `backend`, while the surrounding prose discusses `QUBOTools.backend` (a different function returning a different object). Consider renaming the variable or qualifying the call to avoid conflating the two.

TESTS RUN

- Environment note: Julia 1.10 and 1.11 in this checkout fail to even precompile dependencies (`StaticData not defined` in PrecompileTools; an MOI Benchmarks precompile failure on 1.11) - a local depot/version artifact unrelated to this PR. The PR's documented command uses `julia +release`, which is 1.12.0 here; all runs below are on 1.12.
- Targeted: `test_reformulation_metadata()` -> 25/25 pass.
- Full suite `Pkg.test()` -> all pass (Unit + Integration, 320 integration tests, exit 0).
- Regression check: `QUBOTools.backend` on an uncompiled/empty optimizer still returns a model and does not throw (the new metadata path reads the target objective, but get on the empty target model is safe).

MERGE-READINESS

Mergeable. No blocking issues; the nonblocking items are improvements, not gates. Full test suite is green on the documented Julia version. Because this is my own PR, the formal approval must come from another maintainer.

bernalde (commented):
Re-review of current head `c6227b64a42b71ef3a5cbd889cc44d623212a4d6`.

Note: I am the author of this PR, so GitHub does not permit me to submit APPROVE or REQUEST_CHANGES. This is submitted as COMMENT; any formal approval must come from another maintainer.

Blocking: None.

Nonblocking:
- The existing inline comments from review `4443772787` remain current: serialized `project_original_state` silently projects empty `expansion_terms` to 0, `scale` is asserted as `one(T)`, the JSON-compatible guarantee is structurally tested rather than round-tripped, and `_evaluate_serialized_terms` returns `Int(0)` for an empty term list.

Questions:
- Confirm #132 should stay open after merge for the deferred `is_encoded_consistently` / auxiliary-repair work. The PR correctly uses `Refs #132`, not an auto-closing keyword.
- Consider clarifying the docs example variable named `backend`, since `unsafe_backend(model)` and `QUBOTools.backend(...)` return different objects.

Tests run and outcomes:
- Confirmed local diff totals against the merge-base match GitHub's PR totals: 6 files, +661/-1.
- `julia +release --project=. -e 'using Test; import MathOptInterface as MOI; using ToQUBO; using ToQUBO: Attributes; import QUBOTools; const VI = MOI.VariableIndex; include("test/unit/reformulation.jl"); test_reformulation_metadata()'` passed 25/25.
- `julia +release --project=. -e 'import Pkg; Pkg.test()'` passed 989/989.
- `julia +release --project=docs docs/make.jl` passed; Documenter only warned that deployment was skipped outside CI.
- `julia --project=. -e 'using ToQUBO; println("loaded")'` on local Julia 1.10.11 failed during dependency precompile with a stale local depot issue in PrecompileTools/Parsers/MathOptInterface before reaching PR code. GitHub's Julia 1.10 Ubuntu and Windows jobs are green.
- `julia +release --project=. -e 'import MathOptInterface as MOI; using ToQUBO; import QUBOTools; model = ToQUBO.Optimizer{Float64}(); QUBOTools.backend(model); println("backend ok")'` passed.

Merge-readiness:
- Mergeable from my review. No blocking issues; CI is green; the remaining comments are nonblocking maintainability/docs follow-ups. Formal approval needs another maintainer because this is my own PR.

bernalde:
Review follow-up pushed in commit `cec237d`:

- `cec237dde3ef6f5e465370c85c26087e07cbcc3c` (`fix: address reformulation metadata review comments`)

Main changes:
- Added an explicit `encoded` field for source-variable metadata entries and made serialized `project_original_state` error when an entry is marked unencoded, matching the live-model behavior.
- Added regression coverage for the serialized unencoded-entry path.
- Documented that ToQUBO does not rescale the source objective during reformulation, so recorded `scale` is currently `one(T)`.
- Tightened the structural JSON-compatibility test to reject non-finite floats and documented finite numeric values in the manual.
- Clarified the manual example by naming the MOI optimizer separately from the `QUBOTools` model.

Tests run:
- `git diff --check` passed.
- `julia +release --project=. -e 'using Test; import MathOptInterface as MOI; using ToQUBO; using ToQUBO: Attributes; import QUBOTools; const VI = MOI.VariableIndex; include("test/unit/reformulation.jl"); test_reformulation_metadata()'` passed 29/29.
- `julia +release --project=. -e 'import Pkg; Pkg.test()'` passed 993/993.
- `julia +release --project=docs docs/make.jl` passed; Documenter only warned that deployment was skipped outside CI.

- The non-inline note about `_evaluate_serialized_terms` returning `Int(0)` for an empty term list is left as a follow-up. Unencoded empty expansions now error before evaluation; for a legitimate encoded zero polynomial, the serialized term list has no coefficient from which to infer a better numeric type without expanding the schema.
- `is_encoded_consistently` and exact auxiliary repair remain deferred by design; this PR still uses `Refs #132`, so the issue should stay open for that follow-up scope.

Remaining risks or follow-up:
- GitHub CI passed for `cec237d` after this follow-up was pushed.

src/reformulation.jl
bernalde: Addressed in `cec237d`: source-variable metadata now records `encoded`, and the serialized projection path errors when an entry is marked unencoded. Added a regression test for this case in `test/unit/reformulation.jl`.

src/reformulation.jl
bernalde: Addressed in `cec237d`: `_target_metadata` now documents the invariant that ToQUBO adds penalties/quadratization terms but does not rescale the source objective, so `scale` is currently `one(T)`.

docs/src/manual/3-results.md
bernalde: Addressed in `cec237d`: the manual now states finite numeric values as part of the metadata contract, and the structural compatibility test rejects non-finite floats. I did not add a new JSON dependency.

docs/src/manual/3-results.md
bernalde: Addressed in `cec237d`: the docs now call the JuMP `unsafe_backend(model)` result `optimizer` and qualify the `QUBOTools` conversion as `ToQUBO.QUBOTools.backend(optimizer)`.

PR #170, opened 2026-06-21
Add fast path for already-QUBO models

### LINKED ISSUE #168: Add fast path for already-unconstrained quadratic binary models
## Summary

Add a fast compiler path for source models that are already unconstrained quadratic binary optimization models.

The NPP benchmark is already QUBO-shaped:

- binary variables only;
- no constraints other than binary domains;
- a quadratic objective;
- no reformulation, penalty generation, slack encoding, or constraint elimination should be necessary.

After the recent backend extraction improvements, ToQUBO's remaining NPP gap is no longer caused by `QUBOTools.backend(model)`. The remaining avoidable cost appears to be the ToQUBO `optimize!` / compiler path that still processes an already-QUBO model through general-purpose machinery.

This issue is to implement a direct pass-through / copy path for already-QUBO source models so these cases avoid the full reformulation pipeline.

## Benchmark Evidence

Benchmark context:

- benchmark repo/result set: `ToQUBO-benchmark`;
- local result label: `toqubo-resolved-upstream-2026-06-20`;
- Julia: `1.12.6`;
- ToQUBO.jl: `main` at `daaadececd1027fd33914c734158cd6df6517f55`;
- QUBOTools.jl: `main` at `03861ee8f4c642c43a97df06b7ff2b1a49843daf`;
- QUBOTools version in the benchmark manifest: `0.14.2`.

For NPP at 1,000 variables:

| Package | Total time (s) |
| --- | ---: |
| Amplify | `0.14626216888427734` |
| ToQUBO | `0.523936814` |
| PyQUBO | `0.9405150413513184` |
| dimod | `2.2192113399505615` |

ToQUBO phase breakdown for the same NPP row:

| Phase | Time (s) |
| --- | ---: |
| JuMP model construction | `0.234747696` |
| ToQUBO `optimize!` / compile | `0.277686346` |
| Backend extraction | `0.011494712` |
| Total | `0.523936814` |

Backend extraction is now only about `2.19%` of total runtime for this NPP case. The large remaining ToQUBO-specific target is the `optimize!` / compiler phase.

At the common 660-variable NPP size:

| Package | Total time (s) |
| --- | ---: |
| Amplify | `0.04517412185668945` |
| ToQUBO | `0.153651743` |
| PyQUBO | `0.24483036994934082` |
| dimod | `0.9463310241699219` |
| qubovert | `2.4436628818511963` |
| qiskit | `5.185392141342163` |

TSP is not the motivating regression here. The same benchmark run has ToQUBO performing well on TSP:

| Problem | Variables | Total (s) | JuMP model construction (s) | ToQUBO `optimize!` (s) | Backend extraction (s) |
| --- | ---: | ---: | ---: | ---: | ---: |
| TSP | 10,000 | `2.143814646` | `0.41441261` | `1.616393767` | `0.11300103` |

The NPP case is different because it should not need the general constrained-model compilation path at all.

## Current Behavior

ToQUBO already has a QUBO-shaped model predicate and a special branch:

- `Compiler.compile!(model::Virtual.Model{T}, arch)` checks `is_qubo(model.source_model)`;
- if true, it calls `Compiler.copy!(model, arch)` and returns;
- `is_qubo(model)` is currently:

```julia
is_qubo(model) = is_quadratic(model) && is_unconstrained(model) && is_binary(model) && is_optimization(model)
```

This is the right high-level idea, but the current `Compiler.copy!` path still appears to use the normal encoding / parsing / build path:

- mirror-encodes variables;
- parses the objective into ToQUBO's intermediate representation;
- builds the target model through the general machinery.

For already-unconstrained quadratic binary objectives, that is more work than necessary.

## Proposed Implementation

Implement a direct source-to-target copy path for already-QUBO binary models. This can either replace the current `Compiler.copy!` behavior for `is_qubo(source_model)` or live behind a new helper called from that branch, for example:

```julia
Compiler.copy_qubo_source!(model, arch)
```

The fast path should be intentionally narrow at first:

- source model has only binary variable-domain constraints;
- no non-domain constraints;
- objective is `MOI.VariableIndex`, `MOI.ScalarAffineFunction{T}`, or `MOI.ScalarQuadraticFunction{T}`;
- objective sense is valid and preserved;
- no penalty reformulation or variable encoding is required.

A reasonable first version can target binary QUBO only, because that is what the NPP benchmark exercises. Spin support can be added later if it needs different mapping semantics.

### Suggested Algorithm

For a source model that satisfies the existing `is_qubo(source_model)` predicate:

1. Empty or initialize `model.target_model`.
2. Create target variables in the same order as the source variables.
3. Add `MOI.ZeroOne` constraints for each target variable.
4. Build a source-variable to target-variable index map.
5. Copy the objective sense from the source model to the target model.
6. Read the source objective function.
7. Convert `VariableIndex` and affine objectives into an equivalent scalar quadratic objective if the target expects `SQF{T}`.
8. For `ScalarQuadraticFunction{T}`:
   - remap every quadratic term's variable indices through the source-to-target map;
   - remap every affine term's variable index through the same map;
   - preserve coefficient values and constant term;
   - preserve term ordering only if existing tests rely on it.
9. Set the mapped objective directly on `model.target_model`.
10. Populate only the minimal virtual model metadata needed for existing public APIs to remain coherent.

The key part is to avoid:

- building PBO / PBF intermediates for an objective that is already quadratic;
- creating mirror encodings when source and target variables can be copied directly;
- running penalty or constraint machinery when there are no constraints to reformulate;
- generating large metadata structures unless the caller explicitly requests full metadata.

## Metadata and Public API Requirements

This path should remain compatible with:

- `QUBOTools.backend(model)`;
- `QUBOTools.backend(model; full_metadata=true)`;
- solution projection / variable mapping APIs already supported by ToQUBO;
- objective sense queries;
- variable and constraint queries on the target model.

For a pure pass-through QUBO, full metadata does not need to mimic a full reformulation. It can be minimal "mirror" metadata as long as the public behavior is correct and documented by tests.

If the current metadata contract makes a fully direct bypass risky, an acceptable intermediate implementation is to optimize `Compiler.copy!` itself:

- skip PBO materialization when no encoding or constraints are required;
- skip the general objective parser when the objective is already `ScalarQuadraticFunction{T}`;
- write directly into `target_model`;
- keep whatever metadata fields are required for projection and introspection.

## Tests

Please add targeted tests for the fast path.

Functional tests:

- a model with only `MOI.ZeroOne` variable-domain constraints and a `ScalarQuadraticFunction{T}` objective triggers the fast path;
- the resulting QUBO coefficients match the existing compiler output;
- objective sense is preserved for minimization and maximization;
- variable order is preserved or the source-to-target map is correct and stable;
- affine objectives are accepted and copied correctly;
- single-variable objectives using `MOI.VariableIndex` are accepted and copied correctly;
- constant terms are preserved;
- `QUBOTools.backend(model)` works without full metadata;
- `QUBOTools.backend(model; full_metadata=true)` still works and returns coherent metadata;
- solution projection / variable mapping works for a pass-through QUBO.

Negative / fallback tests:

- models with non-domain constraints still use the normal reformulation path;
- models with continuous, integer, or otherwise non-binary variables do not take the binary fast path;
- models with unsupported objective functions behave as they do today;
- constrained TSP-like models are not incorrectly classified as pass-through QUBOs.

Performance regression coverage:

- add a small benchmark or benchmark test fixture for an NPP-style model:

```julia
@variable(model, x[1:n], Bin)
@objective(model, Min, (sum(s[i] * (2x[i] - 1) for i in 1:n))^2)
```

The CI test should not depend on a hard wall-clock threshold unless the project already has a stable performance-test convention. It is enough to provide a local benchmark that can verify the new path reduces `optimize!` time. If a deterministic internal assertion is possible, assert that the fast-path branch was selected.

## Acceptance Criteria

- Already-unconstrained quadratic binary source models avoid the general reformulation path.
- NPP-style `optimize!` time is substantially reduced from the current `~0.278s` at 1,000 variables on the benchmark machine.
- Backend extraction remains metadata-light by default.
- `full_metadata=true` remains correct for pass-through QUBO models.
- Existing constrained-model behavior is unchanged.
- Existing tests pass.
- The ToQUBO-benchmark NPP result improves without regressing the TSP result.

For reference, the current NPP 1,000-variable lower bound from this benchmark setup is roughly:

```text
JuMP model construction + backend extraction
= 0.234747696 + 0.011494712
= 0.246242408 seconds
```

This issue does not attempt to remove JuMP model-construction time. It targets ToQUBO's compiler / `optimize!` cost for models that are already QUBOs.

## Open Questions

- Should the first implementation support only binary QUBO, or should it also support spin models if they are already present in the source?
- Should the direct-copy helper live in `compiler/compiler.jl`, `compiler/copy.jl`, or closer to the `QUBOModel` implementation?
- What is the smallest metadata payload that keeps `full_metadata=true` and solution projection correct for a pass-through model?
- Should ToQUBO expose an internal flag or debug attribute indicating that the fast path was selected, so this can be tested without relying on timing?

bernalde:
Summary

- Added a direct `Compiler.copy!` path for source models that already satisfy `is_qubo(source_model)`.
- Copies binary variables, objective sense, and `VariableIndex`/affine/quadratic objectives directly into the target QUBO model.
- Preserves mirror-style virtual metadata, compiled objective, compiled Hamiltonian, backend metadata, and solution projection behavior without running constraint, penalty, quadratization, or generic build machinery.
- Added compiler tests for quadratic, affine, and single-variable objectives, min/max sense, metadata/projection, an NPP-style fixture, and constrained-model fallback.

Tests

- `julia --project=. -e 'using Pkg; Pkg.test(; coverage=false)'`
  - Passed: 1408/1408 tests.
- Local warmed NPP-style timing probe, 1000 binary variables, same machine:
  - `origin/main` compilation time: `0.697745332s`
  - this branch compilation time: `0.305040407s`
  - this branch marker: `fast_path=true`
  - target variables: `1000`

Branch Hygiene

- Base branch: `main`
- Source branch point: `origin/main` at `271e1a0`
- Stacked status: not stacked; branch was created directly from current `origin/main`
- Prerequisite PRs: none

Closes #168

### DIFF
--- a/test/unit/compiler/copy.jl
+++ b/test/unit/compiler/copy.jl
@@ -1,0 +1,231 @@
+function _target_objective_pbf(model::ToQUBO.Optimizer{T}) where {T}
+    obj = MOI.get(model.target_model, MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{T}}())
+    f = PBO.PBF{VI,T}()
+
+    for q in obj.quadratic_terms
+        if q.variable_1 == q.variable_2
+            f[q.variable_1] += q.coefficient / 2
+        else
+            f[VI[q.variable_1, q.variable_2]] += q.coefficient
+        end
+    end
+
+    for a in obj.affine_terms
+        f[a.variable] += a.coefficient
+    end
+
+    f[nothing] += obj.constant
+
+    return f
+end
+
+function _compiled_uses_qubo_fast_path(model::ToQUBO.Optimizer)
+    return get(model.compiler_settings, :qubo_fast_path, false) === true
+end
+
+function _set_objective!(model, sense, f)
+    MOI.set(model, MOI.ObjectiveSense(), sense)
+    MOI.set(model, MOI.ObjectiveFunction{typeof(f)}(), f)
+
+    return nothing
+end
+
+function test_compiler_copy()
+    @testset "→ Copy" begin
+        @testset "already-QUBO quadratic objective uses direct copy" begin
+            let model = ToQUBO.Optimizer{Float64}()
+                x, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+                y, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+
+                obj = MOI.ScalarQuadraticFunction{Float64}(
+                    [
+                        MOI.ScalarQuadraticTerm(6.0, x, x),
+                        MOI.ScalarQuadraticTerm(3.0, x, y),
+                        MOI.ScalarQuadraticTerm(5.0, y, x),
+                    ],
+                    [MOI.ScalarAffineTerm(2.0, y)],
+                    7.0,
+                )
+                _set_objective!(model, MOI.MIN_SENSE, obj)
+
+                MOI.optimize!(model)
+
+                x_target = only(MOI.get(model, Attributes.VariableTargetVariables(), x))
+                y_target = only(MOI.get(model, Attributes.VariableTargetVariables(), y))
+                expected = PBO.PBF{VI,Float64}(
+                    7.0,
+                    x_target => 3.0,
+                    y_target => 2.0,
+                    VI[x_target, y_target] => 8.0,
+                )
+
+                @test _compiled_uses_qubo_fast_path(model)
+                @test MOI.get(model.target_model, MOI.ObjectiveSense()) == MOI.MIN_SENSE
+                @test MOI.get(model.target_model, MOI.NumberOfVariables()) == 2
+                @test MOI.get(model.target_model, MOI.NumberOfConstraints{VI,MOI.ZeroOne}()) == 2
+                @test isempty(model.g)
+                @test isempty(model.h)
+                @test isempty(model.s)
+                @test MOI.get(model, Attributes.CompiledObjectiveFunction()) == expected
+                @test MOI.get(model, Attributes.CompiledHamiltonian()) == expected
+                @test _target_objective_pbf(model) == expected
+            end
+        end
+
+        @testset "affine and variable objectives are copied directly" begin
+            let model = ToQUBO.Optimizer{Float64}()
+                x, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+                y, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+
+                obj = MOI.ScalarAffineFunction{Float64}(
+                    [
+                        MOI.ScalarAffineTerm(2.0, x),
+                        MOI.ScalarAffineTerm(-1.0, y),
+                    ],
+                    4.0,
+                )
+                _set_objective!(model, MOI.MAX_SENSE, obj)
+
+                MOI.optimize!(model)
+
+                x_target = only(MOI.get(model, Attributes.VariableTargetVariables(), x))
+                y_target = only(MOI.get(model, Attributes.VariableTargetVariables(), y))
+                expected = PBO.PBF{VI,Float64}(
+                    4.0,
+                    x_target => 2.0,
+                    y_target => -1.0,
+                )
+
+                @test _compiled_uses_qubo_fast_path(model)
+                @test MOI.get(model.target_model, MOI.ObjectiveSense()) == MOI.MAX_SENSE
+                @test MOI.get(model, Attributes.CompiledObjectiveFunction()) == expected
+                @test _target_objective_pbf(model) == expected
+            end
+
+            let model = ToQUBO.Optimizer{Float64}()
+                x, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+                y, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+
+                MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+                MOI.set(model, MOI.ObjectiveFunction{VI}(), x)
+
+                MOI.optimize!(model)
+
+                x_target = only(MOI.get(model, Attributes.VariableTargetVariables(), x))
+                y_target = only(MOI.get(model, Attributes.VariableTargetVariables(), y))
+                expected = PBO.PBF{VI,Float64}(x_target => 1.0)
+
+                @test _compiled_uses_qubo_fast_path(model)
+                @test MOI.get(model, Attributes.CompiledObjectiveFunction()) == expected
+                @test MOI.get(model, Attributes.VariableTargetVariables(), x) == VI[x_target]
+                @test MOI.get(model, Attributes.VariableTargetVariables(), y) == VI[y_target]
+            end
+        end
+
+        @testset "metadata and projection stay coherent for pass-through QUBO" begin
+            let model = ToQUBO.Optimizer{Float64}()
+                x, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+                y, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+
+                obj = MOI.ScalarAffineFunction{Float64}(
+                    [MOI.ScalarAffineTerm(1.0, x)],
+                    0.0,
+                )
+                _set_objective!(model, MOI.MIN_SENSE, obj)
+
+                MOI.optimize!(model)
+
+                backend = QUBOTools.backend(model)
+                @test backend isa QUBOTools.Model
+                @test !haskey(QUBOTools.metadata(backend), "toqubo")
+
+                full_backend = QUBOTools.backend(model; full_metadata = true)
+                metadata = ToQUBO.reformulation_metadata(full_backend)
+
+                @test metadata["source"]["variable_order"] == [x.value, y.value]
+                @test metadata["target"]["num_variables"] == 2
+                @test isempty(metadata["auxiliary_variables"])
+                @test all(entry["role"] == "source" for entry in metadata["target_variables"])
+                @test ToQUBO.original_variables(full_backend) == [x.value, y.value]
+
+                state = [1, 0]
+                live_projection = ToQUBO.project_original_state(model, state)
+                backend_projection = ToQUBO.project_original_state(full_backend, state)
+
+                @test live_projection[x] == 1.0
+                @test live_projection[y] == 0.0
+                @test backend_projection[x.value] == 1.0
+                @test backend_projection[y.value] == 0.0
+            end
+        end
+
+        @testset "NPP-style quadratic fixture selects fast path" begin
+            let model = ToQUBO.Optimizer{Float64}()
+                n = 6
+                weights = collect(1.0:n)
+                total = sum(weights)
+                x = VI[]
+
+                for _ = 1:n
+                    vi, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+                    push!(x, vi)
+                end
+
+                quadratic_terms = MOI.ScalarQuadraticTerm{Float64}[]
+                affine_terms = MOI.ScalarAffineTerm{Float64}[]
+
+                for i = 1:n
+                    push!(quadratic_terms, MOI.ScalarQuadraticTerm(8 * weights[i]^2, x[i], x[i]))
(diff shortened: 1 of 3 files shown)


bernalde (commented):
Blocking issues: none.

Nonblocking issues: none.

Questions: none.

Tests run and outcomes:

- Confirmed PR metadata with `gh pr view`; PR #170 is based on `main`, head `fix/issue-168-qubo-fast-path`, commit `e0d03755cb94dd0462302b4d6d82a3635dd2afa4`.
- Fetched `origin/main` and `origin/fix/issue-168-qubo-fast-path`; local merge-base diff matches PR totals: 3 files changed, 398 additions, 6 deletions.
- Read issue #168 with `gh issue view`; the implementation addresses the requested already-QUBO fast path and keeps backend metadata/projection behavior covered by tests.
- Ran `julia --project=. -e 'using Pkg; Pkg.test(; coverage=false)'`: passed 1408/1408 tests.
- Ran targeted local probes for fast-path compilation, full metadata projection, and repeated optimization: passed.
- Checked CI with `gh pr checks 170`: all reported checks pass.

Merge-readiness:

- I found no actionable issues introduced by this PR. Because I am the PR author, I am submitting this as `COMMENT`; the formal approval/merge decision must come from another maintainer.

test/unit/compiler/copy.jl
```diff
@@ -0,0 +1,231 @@
...
+
+                @test _compiled_uses_qubo_fast_path(model)
+                @test MOI.get(model.target_model, MOI.NumberOfVariables()) == n
+                @test isempty(model.slack)
+                @test isempty(ToQUBO.auxiliary_variables(model))
+                @test QUBOTools.backend(model) isa QUBOTools.Model
+            end
+        end
+
+        @testset "constrained models use normal reformulation path" begin
```
bernalde, line 206: Nonblocking (test gap): this testset covers the constrained-model fallback, but issue #168's negative-test list also asks that a model with a non-binary (integer/continuous) variable not take the binary fast path. Wrongly selecting the fast path there would skip encoding and emit an incorrect QUBO. The gating predicate is covered indirectly in `test/unit/compiler/analysis.jl:74-97` (`!is_binary`/`!is_qubo`), so this is low-risk, but an `optimize!`-level test with e.g. an `MOI.Integer`+`MOI.Interval` variable asserting `!_compiled_uses_qubo_fast_path(model)` and a correct result would close the loop.

bernalde (commented):

I verified the behavior by running it rather than reading alone: full suite green (1408/1408, see below), and I traced the correctness-critical paths (diagonal coefficient convention, Mirror-encoding equivalence, recompilation safety) against the surrounding code.

## 1. Blocking issues

None.

## 2. Nonblocking issues

- Test gap vs the issue's negative-test list. `test/unit/compiler/copy.jl` covers the constrained-model fallback (a `LessThan` constraint), but not the issue's other requested negative case: a model with a non-binary (integer or continuous) variable plus a quadratic objective must go through the general path, not the binary fast path. Wrongly taking the fast path on such a model would skip encoding and emit an incorrect QUBO, so this is worth closing end-to-end. The gating predicate is already covered indirectly — `test/unit/compiler/analysis.jl:74-97` asserts `!is_binary` and `!is_qubo` for a non-binary model — which is why I rate this nonblocking rather than blocking. Suggest adding one `optimize!`-level test with, e.g., an `MOI.Integer`/`MOI.Interval` variable, asserting `!_compiled_uses_qubo_fast_path(model)` and a correct result.

## 3. Questions

- Performance claim not verified. The PR body reports `origin/main` 0.698s vs branch 0.305s on a local NPP probe; issue #168 frames the target as the ~0.278s `optimize!` phase at 1,000 variables. Wall-clock timing is machine-dependent and not reproducible in a review, so I confirmed only functional correctness and the fast-path selection marker (`compiler_settings[:qubo_fast_path]`), not the speedup magnitude. No action needed unless you want a deterministic perf-regression fixture (the issue notes a hard wall-clock threshold should be avoided; the `:qubo_fast_path` assertion already serves as the deterministic signal).

- `Closes #168` is appropriate, not a `Refs` case. The issue scopes a binary-only first version explicitly ("A reasonable first version can target binary QUBO only"), lists spin support as a deferred open question, and asks for a debug marker for fast-path selection — which this PR provides via `compiler_settings[:qubo_fast_path]`. The core ask (direct copy path avoiding the reformulation pipeline for already-QUBO binary models, plus tests) is delivered, so auto-closing on merge is correct.

## 4. Tests run and outcomes

All on a clean checkout of the PR head (`e0d0375`) against merge-base `271e1a0`; diff confirmed at 3 files / +398 / -6, matching the PR totals.

- Full suite: `julia --project=. -e 'using Pkg; Pkg.test()'` — `♡ ToQUBO.jl 0.4.1 Test Suite ♡ | 1408 1408` pass, `Testing ToQUBO tests passed` (exit 0). The new `□ Compiler → Copy` testset is 37/37. Integration (320) including TSP unaffected.
- Could not run the copy testset in isolation via a bare `julia --project=.` because `QUBODrivers` is a test-only dependency resolved only inside `Pkg.test`; ran it through the full harness instead.

Correctness spot-checks beyond the tests:
- Diagonal quadratic handling is correct against MOI's convention: `ScalarQuadraticTerm(c, x, x)` denotes `0.5*c*x^2`, which for binary `x` is `0.5*c*x`, matching `f[term(x)] += c/2` ([src/compiler/compiler.jl](src/compiler/compiler.jl)); off-diagonal `c*x*y` maps to `+= c`. The first copy test exercises this (6.0 diagonal -> 3.0; 3.0+5.0 cross terms -> 8.0).
- The manually constructed `Virtual.Variable{T}(Mirror{T}(), x, VI[y], PBF(y), nothing)` plus `add_constrained_variable(ZeroOne)` and the `source`/`target`/`variables` registration exactly reproduce what `Encoding.encode!` does for a `Mirror` encoding ([src/virtual/encoding.jl:1-28](src/virtual/encoding.jl), [src/encoding/variables/mirror.jl](src/encoding/variables/mirror.jl)), so no mirror-metadata regression.
- Skipping `setup!` is safe: it is effectively a no-op (`src/compiler/setup.jl`), and every `compiler_settings` read uses a defaulted `get`.
- `reset!` at the top of `copy!` is correct and an improvement: `MOI.optimize!` calls `compile!` without a prior reset ([src/wrapper.jl:27-34](src/wrapper.jl)), so re-optimizing a fast-path model now starts from a clean target rather than accumulating state as the old `copy!` could have.

## 5. Merge readiness

Mergeable. The change is well-scoped: the general reformulation path is untouched apart from a harmless `delete!(:qubo_fast_path)` guard, `is_qubo`/`is_binary` gating prevents misclassification, the QUBO coefficients are provably correct, and the full suite passes. The only follow-up worth doing is the non-binary fallback test (nonblocking). I would be comfortable with this merging once another maintainer records the formal approval.

bernalde:
Addressed review feedback in `023dc3c1e837e2413ce7b30ec6f464e147a7f2f3`.

Commits pushed:

- `023dc3c` test: cover non-binary QUBO fallback (#170)

Main changes made:

- Added an `optimize!`-level test for a binary variable plus an `MOI.Integer` + `MOI.Interval` variable with a quadratic objective.
- The test asserts the model does not satisfy `is_qubo`, does not select `:qubo_fast_path`, compiles through the normal reformulation path, preserves the expected integer projection, and still supports `QUBOTools.backend(model)`.

Tests run and results:

- Focused non-binary fallback probe: passed.
- `julia --project=. -e 'using Pkg; Pkg.test(; coverage=false)'`: passed 1415/1415 tests.
- `gh pr checks 170` after push: label passed; Julia and docs/build checks are pending on the new commit.

- None.

Remaining risks or follow-up items:

- None identified. The CI matrix is still running for the pushed commit.

test/unit/compiler/copy.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/023dc3c1e837e2413ce7b30ec6f464e147a7f2f3. I added an `optimize!`-level non-binary domain fallback test in `test/unit/compiler/copy.jl` that asserts the normal path is used, projection is coherent, and backend extraction still works. The focused probe and full suite passed locally.

Issue #162, opened 2026-06-20
Track QOBLib routing penalty-scaling divergence


bernalde:
PR work for #153's QOBLIB routing pilot found a large gap between ToQUBO-generated QUBO coefficient ranges and the pinned QOBLIB canonical QS metrics row for XSH-n20-k4-01.

Observed in the pilot report:

- QOBLIB QS metrics row: min coefficient -1.2874612219171257e10, max coefficient 7.397605618245156e9.
- ToQUBO native range: min coefficient about -1.4036904e13, max coefficient about 2.7938839e13.
- ToQUBO QOBLIB-style symmetrized range: min coefficient about -1.4036904e13, max coefficient about 1.5646507e13.
- Objective offset is about 1.5412343e13, while the known rounded CVRPLIB solution cost is 646 and the source-model sqrt objective projection is about 646.67.

The source transcription and known-incumbent feasibility checks pass, so this appears to be a penalty-scaling or conversion-convention divergence rather than a basic source-model transcription error.

Follow-up work should determine whether this is expected under the current penalty heuristic, whether routing-specific scaling is needed, and whether the canonical QOBLIB conversion uses different treatment for the MTZ load variables, capacity bounds, or redundant capacity constraints.

Related: #145, #153.

PR #173, opened 2026-06-21
Add objective-range penalty policy

### LINKED ISSUE #172: Add objective-range exact penalty policy for tighter reformulations
## Context

PR #171 diagnosed the QOBLib `network05` penalty-scaling divergence from #160 in a way that points to a concrete improvement opportunity in ToQUBO's automatic penalty inference.

Key observations from #171:

- QOBLib's `network05` QS metrics were reproduced exactly from the local QOBLib LP using Qiskit's default converter path:
  - Python 3.12
  - Qiskit 2.4.2
  - qiskit-optimization 0.7.0
  - gurobipy 13.0.2
- Qiskit `QuadraticProgramToQubo()` used a uniform converter penalty of `1000001.0`.
- That value comes from Qiskit's default rule:

```text
penalty = 1 + (objective upper bound - objective lower bound)
```

- For `network05`, the source objective is `min z` with `0 <= z <= 1000000`, so the objective range is `1000000` and the sufficient exact penalty is `1000001`.
- The reproduced QOBLib metrics matched the pinned row exactly:
  - variables: `3640`
  - density: `0.05271012974940467`
  - min coefficient: `-4.75713475713e17`
  - max coefficient: `4.5260616934376417e18`
- ToQUBO's current default heuristic produced much larger penalties and coefficients for the same source transcription:
  - largest applied ToQUBO penalty family: `flow_balance`
  - largest applied ToQUBO penalty: about `5.2338628e14`
  - ToQUBO QOBLIB-style coefficient range: about `[-1.18e26, 1.19e26]`

This does not mean Qiskit's QUBO is numerically small; `~1e18` coefficients are still large. But it shows that ToQUBO's current automatic penalties can be orders of magnitude larger than a standard sufficient exact-penalty bound for bounded integer models.

Related issues and PRs:

- #145
- #160
- #162
- #171

## Mathematical basis

For a minimization problem with bounded original objective

```text
L <= f(x) <= U
```

and penalty functions `p_i(x)` such that:

- `p_i(x) = 0` when the corresponding modeled condition is satisfied;
- `p_i(x) > 0` when the condition is violated;
- the minimum positive violation gap is `epsilon_i > 0`;

then an exact-penalty sufficient condition is:

```text
rho_i * epsilon_i > U - L
```

Equivalently:

```text
rho_i > (U - L) / epsilon_i
```

For squared equality residuals with integral coefficients and integral encoded variables, any nonzero residual has absolute value at least `1`, so `epsilon_i >= 1`. In that common case, the uniform Qiskit-style value:

```text
rho = (U - L) + 1
```

is sufficient in exact arithmetic, assuming the original feasible set is nonempty.

For generic ToQUBO penalty PBFs, using each penalty function's own minimum positive gap is more general:

```text
rho_i = (U - L + margin) / epsilon_i
```

where `epsilon_i` can be obtained from the generated penalty function, e.g. via `PBO.mingap(g)`, `PBO.mingap(h)`, or `PBO.mingap(s)`.

This is very close in shape to the current ToQUBO formula, but the critical difference is the objective range bound used for the numerator.

Current implementation:

```julia
δ = PBO.maxgap(model.f)
ρ = scale * sign * (δ / ϵ + offset)
```

Relevant file:

- `src/compiler/penalties.jl`

The QOBLib network case suggests that `PBO.maxgap(model.f)` is not giving the tight global objective range bound needed here, or at least is not behaving like Qiskit's objective upper-minus-lower bound for this reformulation. This issue should investigate that distinction and add a policy based on a mathematically interpretable objective range.

## Proposed work

Add an automatic penalty policy based on an objective range exact-penalty bound, and make it the preferred/default policy whenever ToQUBO can certify the required conditions.

This issue should not treat backward compatibility as a reason to keep the weaker heuristic as the default for models where we have a mathematically justified tighter bound. The current heuristic should remain available as an explicit legacy policy and as a fallback for cases where the objective range, penalty nonnegativity, or positive violation gap cannot be certified. For certifiable bounded reformulations, the objective-range policy should be the target default because it is smaller, interpretable, and supported by an exact-penalty argument.

Possible API shape:

```julia
MOI.set(model, ToQUBO.Attributes.PenaltyPolicy(), ToQUBO.Attributes.ObjectiveRangePenalty())
```

or:

```julia
MOI.set(model, ToQUBO.Attributes.AutomaticPenaltyPolicy(), ToQUBO.Attributes.ObjectiveRangePenalty(; margin = 1))
```

The naming is intentionally open. The important semantic distinction is:

- current policy: gap-ratio heuristic using `PBO.maxgap(model.f)`;
- new default-when-certifiable policy: exact-penalty bound using an objective lower/upper range and each generated penalty function's violation gap.

## Extracting the objective range from JuMP/MOI models

ToQUBO receives models through MOI, including JuMP models copied into `ToQUBO.Optimizer`. The bound extraction should be implemented at the MOI/ToQUBO compiler layer, not by depending on JuMP internals.

Candidate sources for the range:

1. Source MOI objective plus source variable bounds.
2. The compiled pseudo-boolean objective `model.f` after ToQUBO variable encoding, before penalties.

Option 2 may be more robust because ToQUBO already has encoded the source variables into target binary variables and already stores the compiled objective as a PBF. If `PBO.bounds(model.f)` returns valid lower/upper bounds for the compiled objective, then:

```julia
L, U = PBO.bounds(model.f)
range = U - L
```

This also handles objective constants and the actual encoded coefficients in the representation that penalties will be added to.

Still, the issue should evaluate both approaches.

For direct MOI source extraction, the implementation needs to inspect:

- objective sense via `MOI.ObjectiveSense`;
- objective function type via `MOI.ObjectiveFunctionType`;
- `MOI.VariableIndex`;
- `MOI.ScalarAffineFunction`;
- `MOI.ScalarQuadraticFunction`;
- source variable bounds from MOI constraints:
  - `MOI.ZeroOne`;
  - `MOI.Integer`;
  - `MOI.Interval`;
  - `MOI.LessThan`;
  - `MOI.GreaterThan`;
  - possibly `MOI.EqualTo` on a variable.

For affine objectives, a finite interval bound is straightforward:

```text
constant + sum_i c_i * (lower_i if c_i >= 0 else upper_i)
constant + sum_i c_i * (upper_i if c_i >= 0 else lower_i)
```

For quadratic objectives, exact global bounds are harder. Options:

- use the compiled PBF and `PBO.bounds(model.f)` after encoding;
- for simple interval fallback, bound each product term over endpoint combinations;
- when bounds are not finite or the result is not certifiable, fall back to the current heuristic and record that fallback in metadata.

## Penalty application details

For each generated penalty function:

- constraint penalties: `model.g`
- variable-encoding penalties: `model.h`
- slack-variable penalties: `model.s`

compute:

```julia
epsilon = PBO.mingap(penalty_function)
rho = sign * scale * ((objective_range + margin) / epsilon)
```

where:

- `sign` follows the existing objective-sense convention;
- `scale` and constraint-specific scale overrides should still apply;
- explicit penalty hints must continue to take precedence;
- `LinearPenalty()` should probably continue to require an explicit hint unless we can prove exactness for signed linear penalties separately.

Open design question: should `PenaltyOffset` remain additive under this policy, or should `margin` replace it? The current offset is inside `δ / ϵ + β`; an objective-range policy may want:

```text
(range + margin) / epsilon
```

rather than:

```text
range / epsilon + offset
```

Those differ when `epsilon != 1`.

## Metadata and diagnostics

Expose enough information to verify and debug the policy:

- automatic penalty policy name;
- objective lower bound;
- objective upper bound;
- objective range;
- objective range source:
  - source MOI affine/quadratic interval;
  - compiled PBF bounds;
  - fallback;
- margin;
- per-penalty `epsilon`;
- final inferred penalty;
- fallback reason if objective range could not be certified.

The metadata should be visible through existing reformulation metadata APIs so benchmark pilots can report it.

Relevant existing metadata surface:

- `src/reformulation.jl`
- applied penalties metadata
- compiled objective metadata

## Acceptance criteria

- Add an objective-range exact-penalty policy attribute.
- Use the objective-range policy by default when the compiler can certify finite objective bounds, nonnegative penalty functions, and positive violation gaps.
- Keep explicit penalty hints working exactly as they do today.
- Keep the current heuristic available as an explicit legacy/fallback policy, not as the preferred behavior for certifiable bounded models.
- Record fallback reasons when the objective-range policy cannot be certified and the compiler uses the legacy heuristic.
- Add unit tests for:
  - affine bounded minimization objective;
  - affine bounded maximization objective;
  - integer equality residual with `epsilon = 1`;
  - generated penalty function with `epsilon != 1`;
  - unbounded objective fallback;
  - explicit hint precedence;
  - `PenaltyScale` and per-constraint scale interaction;
  - `LinearPenalty()` behavior.
- Add or update QOBLib benchmark diagnostics:
  - `network05` should report the objective range `1000000` and Qiskit-style sufficient penalty `1000001` under the new policy.
  - Re-run the network pilot to show how much the coefficient range tightens compared with the current default.
  - Re-run the routing pilot from #162 to see whether the same policy explains or reduces that divergence.
- Update the documentation and release notes:
  - explain the exact-penalty proof and the condition `rho_i * epsilon_i > U - L`;
  - document how objective bounds are extracted from MOI/JuMP-backed models;
  - document when ToQUBO falls back to the legacy heuristic;
  - document how explicit penalty hints, `PenaltyScale`, per-constraint scale overrides, and any margin/offset setting interact with the new policy;
  - add an example showing the `network05` objective range `1000000` and inferred penalty `1000001`.

## Important caveats

- The objective-range bound is a sufficient exactness condition, not a numerically optimal penalty.
- It assumes finite objective bounds over the encoded domain.
- It assumes the generated penalty functions are nonnegative with known positive violation gaps.
- For floating-point, non-integral, or ill-scaled constraints, the proof may not hold as cleanly. The implementation should detect non-certifiable cases and fall back rather than silently claiming exactness.
- Matching Qiskit/QOBLib is useful for benchmark parity, but the larger goal is tighter and more interpretable penalties in ToQUBO reformulations.

bernalde:
## Summary
- Add an objective-range exact-penalty policy as the default automatic penalty inference path.
- Keep the legacy maxgap heuristic available through `LegacyPenalty` and use it as the recorded fallback when objective-range inference cannot be certified.
- Expose `PenaltyPolicy` and `PenaltyPolicyMetadata`, including JuMP-facing get/set and metadata retrieval.
- Update QOBLib network/routing diagnostics, documentation, release notes, and golden tests for the tighter default penalties.

## Validation
- `julia --project=. -e 'using Test; import MathOptInterface as MOI; import ToQUBO; using ToQUBO: Attributes, Encoding; include("test/unit/compiler/penalties.jl"); test_compiler_penalties()'` (28/28)
- `julia --project=. -e 'import Pkg; Pkg.test()'` (1500/1500)
- `julia --project=docs docs/make.jl --skip-deploy`
- `git diff --check`

## Branch Hygiene
- Base branch: `main`
- Source branch point: `origin/main` at `b1406fea2db989bc9137569b7f2d2252da6f7b39`
- Stacked status: not stacked
- Prerequisite PRs: none

Closes #172.

### DIFF
--- a/benchmarks/qoblib/network_pilot.jl
+++ b/benchmarks/qoblib/network_pilot.jl
@@ -57,7 +57,6 @@ const PROVENANCE = Dict{String,Any}(
     "canonical_qubo_artifact_available_at_commit" => false,
     "canonical_qubo_artifact_note" =>
         "The pinned QOBLIB tree contains the network05.qs metrics row but no stored network05.qs or network05.qs.xz artifact, so this pilot compares against the metrics table row.",
-    "penalty_scaling_follow_up" => "https://github.com/JuliaQUBO/ToQUBO.jl/issues/172",
 )
 
 const UPSTREAM_VERIFICATION = Dict{String,Any}(
@@ -836,7 +835,6 @@ function write_markdown_report(io::IO, report::AbstractDict)
     write(io, "- Canonical QUBO artifact note: $(provenance["canonical_qubo_artifact_note"])\n")
     write(io, "- Model license: $(provenance["qoblib_model_license"])\n")
     write(io, "- Data license: $(provenance["qoblib_data_license"])\n")
-    write(io, "- Penalty scaling follow-up: $(provenance["penalty_scaling_follow_up"])\n")
     write(io, "- Generated collection label: $(provenance["collection"])\n\n")
 
     write(io, "## Upstream Verification\n\n")
--- a/benchmarks/qoblib/network_pilot.jl
+++ b/benchmarks/qoblib/network_pilot.jl
@@ -57,7 +57,7 @@ const PROVENANCE = Dict{String,Any}(
     "canonical_qubo_artifact_available_at_commit" => false,
     "canonical_qubo_artifact_note" =>
         "The pinned QOBLIB tree contains the network05.qs metrics row but no stored network05.qs or network05.qs.xz artifact, so this pilot compares against the metrics table row.",
-    "penalty_scaling_follow_up" => "https://github.com/JuliaQUBO/ToQUBO.jl/issues/160",
+    "penalty_scaling_follow_up" => "https://github.com/JuliaQUBO/ToQUBO.jl/issues/172",
 )
 
 const UPSTREAM_VERIFICATION = Dict{String,Any}(
@@ -530,6 +530,7 @@ function _constraint_penalty_diagnostics(
     metadata::AbstractDict,
 )
     expansion_coefficients = _source_variable_expansion_coefficients(metadata)
+    penalty_policy = metadata["penalty_policy"]
     families = Dict{String,Any}[]
 
     for category in CONSTRAINT_CATEGORY_ORDER
@@ -590,7 +591,10 @@ function _constraint_penalty_diagnostics(
     ])]
 
     return Dict{String,Any}(
-        "heuristic" => "scale * sigma * (delta / epsilon + beta)",
+        "heuristic" => "objective-range exact penalty with legacy fallback",
+        "automatic_penalty_policy" => penalty_policy["policy"],
+        "objective_range" => penalty_policy["objective_bounds"]["range"],
+        "fallback_count" => penalty_policy["fallback_count"],
         "default_penalty_scale" => MOI.get(model, Attributes.PenaltyScale()),
         "default_penalty_offset" => MOI.get(model, Attributes.PenaltyOffset()),
         "constraint_families" => families,
@@ -645,7 +649,7 @@ function _scaling_diagnostics(target::AbstractDict, penalty_diagnostics::Abstrac
         "largest_penalty_times_expanded_residual_coefficient_squared" =>
             penalty_diagnostics["largest_expanded_residual_scale"],
         "assessment" =>
-            "The source transcription and incumbent checks pass. QOBLIB's network05 QS metrics are reproduced by Qiskit's default converter with a uniform penalty of 1000001.0. Under ToQUBO's default automatic penalty heuristic, flow-balance constraints dominate after integer flow-variable expansion; the largest applied penalty and expanded residual coefficient from that family reproduce ToQUBO's larger coefficient scale. Matching the pinned QS metrics row requires a benchmark-compatible penalty policy, not a source model transcription change.",
+            "The source transcription and incumbent checks pass. QOBLIB's network05 QS metrics are reproduced by Qiskit's default converter with a uniform penalty of 1000001.0. ToQUBO's objective-range exact-penalty policy now infers the same uniform applied penalty for all network constraints. The remaining coefficient gap is driven by residual expansion and encoding differences, with arc-linking dominating after the 1000000 source coefficient is expanded into the QUBO.",
     )
 end
 
@@ -768,7 +772,7 @@ function run_network_pilot()
         "known_incumbent" => _known_incumbent_summary(),
         "comparison" => _comparison(target),
         "follow_up" =>
-            "The pilot found a major coefficient-scaling gap: ToQUBO's generated network QUBO coefficient range remains about eight orders of magnitude larger than the pinned QOBLIB QS metrics row even under the QOBLIB-style symmetrized convention. The source transcription and incumbent feasibility checks pass, and the local QOBLIB converter reproduces the pinned QS metrics row with Qiskit's default uniform penalty of 1000001.0. This points to a penalty-policy difference rather than a bad QOBLIB conversion; issue https://github.com/JuliaQUBO/ToQUBO.jl/issues/160 tracks a benchmark-compatible penalty policy before expanding the network benchmark class.",
+            "The objective-range exact-penalty policy tightens ToQUBO's applied network penalties from many family-dependent values up to about 5.23e14 down to the Qiskit/QOBLib-compatible uniform value 1000001.0. The source transcription and incumbent feasibility checks pass, and the local QOBLIB converter reproduces the pinned QS metrics row. The remaining coefficient-range gap is no longer caused by oversized applied penalties; it should be interpreted alongside ToQUBO's variable/slack encodings, residual expansion coefficients, and the target-variable delta before expanding the network benchmark class.",
     )
 end
 
@@ -1021,6 +1025,18 @@ function write_markdown_report(io::IO, report::AbstractDict)
     write(
         io,
         "- Automatic penalty heuristic: `$(penalty_diagnostics["heuristic"])`\n",
+    )
+    write(
+        io,
+        "- Automatic penalty policy: `$(penalty_diagnostics["automatic_penalty_policy"])`\n",
+    )
+    write(
+        io,
+        "- Objective range used for automatic penalties: $(_fmt(penalty_diagnostics["objective_range"]))\n",
+    )
+    write(
+        io,
+        "- Penalty policy fallback count: $(penalty_diagnostics["fallback_count"])\n",
     )
     write(
         io,
--- a/docs/src/manual/4-settings.md
+++ b/docs/src/manual/4-settings.md
@@ -54,9 +54,19 @@ where `s` is [`ToQUBO.Attributes.PenaltyScale`](@ref), `β` is
 [`ToQUBO.Attributes.PenaltyOffset`](@ref), `σ` is `1` for minimization models
 and `-1` for maximization models, `U - L` is the compiled pseudo-boolean
 objective range, and `ϵ` is the smallest positive value of the generated
-nonnegative penalty function. With `β > 0`, any infeasible assignment is
-penalized by more than the largest possible objective improvement available
-within the compiled objective range.
+nonnegative penalty function. With the default scale `s = 1` and `β > 0`, any
+infeasible assignment is penalized by more than the largest possible objective
+improvement available within the compiled objective range. With custom scales,
+the same sufficient exactness condition is `s * (U - L + β) > U - L`; smaller
+scales can make the penalty a tuning heuristic rather than a certified exact
+penalty.
+
+The objective range is computed from interval bounds on the compiled PBF over
+encoded target binary variables. For normal finite JuMP/MOI models this bound
+is finite after encoding, but it may overestimate the exact source-objective
+span when an encoding has redundant target states or feasibility relations such
+as one-hot sums. That overestimate keeps the penalty sufficient, but it is not
+always the tightest possible objective range.
 
 When the objective range or positive penalty gap cannot be certified, ToQUBO
 falls back for that coefficient to [`ToQUBO.Attributes.LegacyPenalty`](@ref):
@@ -66,9 +76,10 @@ falls back for that coefficient to [`ToQUBO.Attributes.LegacyPenalty`](@ref):
 ```
 
 where `δ` is the historical objective gap estimate from `PBO.maxgap`.
-The selected policy, objective bounds, and any fallback reasons are exposed
-through [`ToQUBO.Attributes.PenaltyPolicyMetadata`](@ref) and the
-`"penalty_policy"` field in reformulation metadata.
+The selected policy, objective bounds, inferred penalties with their `ϵ`
+sources, and any fallback reasons are exposed through
+[`ToQUBO.Attributes.PenaltyPolicyMetadata`](@ref) and the `"penalty_policy"`
+field in reformulation metadata.
 
 The precedence is:
 
--- a/docs/src/manual/4-settings.md
+++ b/docs/src/manual/4-settings.md
@@ -39,19 +39,36 @@ ToQUBO.Attributes.StableQuadratization
 
 ## Variable & Constraint Encoding
 
-### Penalty Heuristic
+### Automatic Penalty Inference
 
 When a constraint, variable encoding, or slack-variable encoding needs a penalty
-coefficient and no explicit hint is set, ToQUBO uses the automatic heuristic
+coefficient and no explicit hint is set, ToQUBO infers one automatically. The
+default [`ToQUBO.Attributes.ObjectiveRangePenalty`](@ref) policy uses an
+objective-range exact-penalty bound:
+
+```math
+\rho = s \cdot \sigma \cdot \left(\frac{U - L + \beta}{\epsilon}\right)
+```
+
+where `s` is [`ToQUBO.Attributes.PenaltyScale`](@ref), `β` is
+[`ToQUBO.Attributes.PenaltyOffset`](@ref), `σ` is `1` for minimization models
+and `-1` for maximization models, `U - L` is the compiled pseudo-boolean
+objective range, and `ϵ` is the smallest positive value of the generated
+nonnegative penalty function. With `β > 0`, any infeasible assignment is
+penalized by more than the largest possible objective improvement available
+within the compiled objective range.
+
+When the objective range or positive penalty gap cannot be certified, ToQUBO
+falls back for that coefficient to [`ToQUBO.Attributes.LegacyPenalty`](@ref):
 
 ```math
 \rho = s \cdot \sigma \cdot \left(\frac{\delta}{\epsilon} + \beta\right)
 ```
 
-where `s` is [`ToQUBO.Attributes.PenaltyScale`](@ref), `β` is
-[`ToQUBO.Attributes.PenaltyOffset`](@ref), `σ` is `1` for minimization models
-and `-1` for maximization models, `δ` is the objective gap estimate, and `ϵ` is
-the smallest positive gap in the generated penalty function.
+where `δ` is the historical objective gap estimate from `PBO.maxgap`.
+The selected policy, objective bounds, and any fallback reasons are exposed
+through [`ToQUBO.Attributes.PenaltyPolicyMetadata`](@ref) and the
+`"penalty_policy"` field in reformulation metadata.
 
 The precedence is:
 
@@ -62,20 +79,25 @@ The precedence is:
    [`ToQUBO.Attributes.ConstraintPenaltyOffset`](@ref) apply to the source
    constraint and to the slack-variable encoding penalty generated for that
    constraint.
-3. Global [`ToQUBO.Attributes.PenaltyScale`](@ref) and
+3. Global [`ToQUBO.Attributes.PenaltyPolicy`](@ref),
+   [`ToQUBO.Attributes.PenaltyScale`](@ref), and
    [`ToQUBO.Attributes.PenaltyOffset`](@ref) apply everywhere else.
 
-The defaults are `PenaltyScale() == 1.0` and `PenaltyOffset() == 1.0`, matching
-the previous heuristic. The coefficient is positive for minimization models and
-negative for maximization models, so explicit hints should use the same sign as
-the coefficient you want the compiler to apply.
+The defaults are `PenaltyPolicy() == ObjectiveRangePenalty()`,
+`PenaltyScale() == 1.0`, and `PenaltyOffset() == 1.0`. The coefficient is
+positive for minimization models and negative for maximization models, so
+explicit hints should use the same sign as the coefficient you want the
+compiler to apply.
 
 Use the controls according to how much of the model you want to change:
 
+- Use [`ToQUBO.Attributes.PenaltyPolicy`](@ref) to select the automatic
+  inference policy. Set it to [`ToQUBO.Attributes.LegacyPenalty`](@ref) only
+  when you need to reproduce the historical maxgap-based coefficients.
 - Use [`ToQUBO.Attributes.PenaltyScale`](@ref) and
   [`ToQUBO.Attributes.PenaltyOffset`](@ref) to change all automatically
-  inferred penalties while preserving their relative dependence on the
-  generated objective and penalty gaps.
+  inferred penalties while preserving their policy-dependent objective and
+  penalty-gap dependence.
 - Use [`ToQUBO.Attributes.ConstraintPenaltyScale`](@ref) and
   [`ToQUBO.Attributes.ConstraintPenaltyOffset`](@ref) when one source
   constraint needs a different automatic penalty. These overrides also apply to
@@ -94,10 +116,11 @@ Use the controls according to how much of the model you want to change:
 
 ### Changing Penalty Values
 
-For feasibility tuning, start by sweeping `PenaltyScale()` while keeping
-explicit hints unset. If only one constraint needs adjustment, override that
-constraint's automatic scale or offset. Pin exact penalty values only when you
-have a known coefficient to apply.
+For feasibility tuning, start by inspecting
+[`ToQUBO.Attributes.PenaltyPolicyMetadata`](@ref) and sweeping `PenaltyScale()`
+while keeping explicit hints unset. If only one constraint needs adjustment,
+override that constraint's automatic scale or offset. Pin exact penalty values
+only when you have a known coefficient to apply.
 
 ```@example penalty-settings
 using JuMP
@@ -135,17 +158,19 @@ rho_capacity = get_attribute(capacity, Attributes.ConstraintEncodingPenalty())
 rho_assignment = get_attribute(assignment, Attributes.ConstraintEncodingPenalty())
 eta_capacity = get_attribute(capacity, Attributes.SlackVariableEncodingPenalty())
 theta_z = get_attribute(z, Attributes.VariableEncodingPenalty())
+policy_metadata = get_attribute(model, Attributes.PenaltyPolicyMetadata())
 
-# The capacity penalty is inferred by the automatic heuristic.
+# The capacity penalty is inferred by the automatic policy.
 @assert rho_capacity !== nothing
 @assert rho_assignment == 12.0
 @assert eta_capacity == 9.0
 @assert theta_z == 6.0
+@assert policy_metadata["policy"] == "ObjectiveRangePenalty"
 ```
 
-Very small `ϵ` values can produce large penalties. That usually means the
-penalty function has nearly tied infeasible states, so treat the heuristic as a
-starting point and compare feasibility across several scales.
+Very small `ϵ` values or broad objective ranges can produce large penalties.
+Inspect the policy metadata before overriding the policy or pinning explicit
+penalties.
 
 ### Constraint Penalty Methods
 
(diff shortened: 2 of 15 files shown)


docs/src/manual/4-settings.md
```diff
@@ -39,19 +39,36 @@ ToQUBO.Attributes.StableQuadratization
...
+\rho = s \cdot \sigma \cdot \left(\frac{U - L + \beta}{\epsilon}\right)
 ```
 
 where `s` is [`ToQUBO.Attributes.PenaltyScale`](@ref), `β` is
 [`ToQUBO.Attributes.PenaltyOffset`](@ref), `σ` is `1` for minimization models
-and `-1` for maximization models, `δ` is the objective gap estimate, and `ϵ` is
-the smallest positive gap in the generated penalty function.
+and `-1` for maximization models, `U - L` is the compiled pseudo-boolean
+objective range, and `ϵ` is the smallest positive value of the generated
+nonnegative penalty function. With `β > 0`, any infeasible assignment is
```
bernalde: Nonblocking: please qualify this exactness statement with the scale condition. `β > 0` is sufficient for the default `PenaltyScale() == 1`, and more generally exactness needs `s * (U - L + β) > U - L`. With a smaller scale, the stated guarantee no longer follows.

bernalde (commented):
Formal verdict requires another maintainer because I am the PR author. I would approve from this review pass; the note below is nonblocking.

Blocking issues: none.

Nonblocking issues:
- The exactness proof text should qualify the role of `PenaltyScale()`. As written, `β > 0` is stated as sufficient, but the actual coefficient is multiplied by `s`. For `0 < s < (U - L) / (U - L + β)`, an infeasible assignment can still be cheaper/better than the best feasible assignment, so the documentation should state either the default-scale condition or the general inequality.

Questions: none.

Tests run and outcomes:
- `julia --project=. -e 'using Test; import MathOptInterface as MOI; import ToQUBO; using ToQUBO: Attributes, Encoding; include("test/unit/compiler/penalties.jl"); test_compiler_penalties()'`: passed, 28/28.
- `julia --project=. -e 'import Pkg; Pkg.test()'`: passed, 1500/1500.
- `julia --project=docs docs/make.jl --skip-deploy`: passed.
- `gh pr checks 173`: all reported checks passed.

Merge-readiness: no blocking issues found in this pass; another maintainer can provide the formal approval.

src/compiler/penalties.jl
```diff
@@ -3,15 +3,120 @@ function _uses_linear_equality_penalty(model::Virtual.Model, ci::CI)::Bool
...
 
-function _inferred_penalty_factor(sign, δ, ϵ, scale, offset)
+function _legacy_penalty_factor(sign, δ, ϵ, scale, offset)
     return scale * sign * (δ / ϵ + offset)
 end
 
+function _objective_range_metadata(model::Virtual.Model)
+    lower, upper = PBO.bounds(model.f)
+    range = upper - lower
+    finite = isfinite(lower) && isfinite(upper) && isfinite(range) && range >= zero(range)
```
bernalde, line 13: Two notes on this certification. (1) Because `model.f` is the compiled PBF over already-encoded target binaries, `PBO.bounds` is always finite, so this `finite` flag is effectively always true and the `!objective_bounds["finite"]` fallback branch is unreachable in practice. Issue #172's required "unbounded objective fallback" test is therefore absent and arguably not writable against this approach. Not a correctness bug, but either document that the compiled-PBF range is always finite (fallback reserved for degenerate `epsilon`) or make the path reachable and test it. (2) `PBO.bounds(model.f)` ignores encoding feasibility: for an `Integer` on `[0,3]` with `OneHot()` and objective `x` I measured `range = 6`, not the true `3`. Safe (over-estimate keeps `rho*epsilon > U-L`), but looser than the true objective span under one-hot/unary encodings; worth a doc note.

src/compiler/penalties.jl
```diff
@@ -25,13 +130,17 @@ function penalties!(model::Virtual.Model{T}, ::AbstractArchitecture) where {T}
                 )
             end
 
-            ϵ = PBO.mingap(g)
+            ϵ, _gap_source = _positive_penalty_gap(g)
```
bernalde: Nonblocking: `_gap_source` is destructured here (and at lines 154 and 175) but never used. Issue #172 asks for per-penalty `epsilon` and its source in the policy metadata; today only objective bounds and fallbacks are recorded. Either store `epsilon`/`_gap_source` per penalty in `context`, or drop the unused binding. Also note the `epsilon != 1` branch of `_positive_penalty_gap` (the `PBO.mingap` path) has no test — all current tests hit the integer-valued `epsilon = 1` shortcut.

benchmarks/qoblib/network_pilot.jl
```diff
@@ -57,7 +57,7 @@ const PROVENANCE = Dict{String,Any}(
     "canonical_qubo_artifact_available_at_commit" => false,
     "canonical_qubo_artifact_note" =>
         "The pinned QOBLIB tree contains the network05.qs metrics row but no stored network05.qs or network05.qs.xz artifact, so this pilot compares against the metrics table row.",
-    "penalty_scaling_follow_up" => "https://github.com/JuliaQUBO/ToQUBO.jl/issues/160",
+    "penalty_scaling_follow_up" => "https://github.com/JuliaQUBO/ToQUBO.jl/issues/172",
```
bernalde: Question: this repoints `penalty_scaling_follow_up` to #172, the issue this PR closes. Once merged, the network pilot will reference a closed issue as a "follow-up." Point it at an open tracker (or remove it), as the routing pilot does by keeping #162.

bernalde (commented):

I verified the behavior by running it: full suite green (1500/1500), and I ran a standalone probe of the compiled-objective bounds to check the certification/fallback logic rather than reading alone. The exact-penalty math is sound: `PBO.bounds(model.f)` returns interval-arithmetic bounds (sum of negative/positive coefficients), so the inferred range over-approximates the true achievable objective range, which keeps `rho*epsilon > U-L` sufficient. The new default is therefore never less safe than required, and the network05 case reaches the issue's target (range 1000000, penalty 1000001).

## 1. Blocking issues

None.

## 2. Nonblocking issues

- Tightness caveat: the objective range is taken from the compiled PBF over target binaries, which ignores encoding feasibility (e.g. one-hot `sum == 1`). My probe of an `Integer` variable on `[0,3]` with `OneHot()` and objective `x` reported `range = 6`, not the true `3` (`f = y1 + 2*y2 + 3*y3`, bounds `[0,6]`). This stays exact (over-estimate), but for one-hot/unary encodings the inferred penalty is looser than the true objective span, partially undercutting the "tighter, interpretable" goal. Worth a sentence in `docs/src/manual/4-settings.md` noting the range is the compiled-PBF interval bound and can exceed the source objective span under redundant encodings.

- Unused binding / partial metadata: `_gap_source` is destructured at three sites ([src/compiler/penalties.jl:133,154,175](src/compiler/penalties.jl)) and never used. Issue #172's metadata wishlist asks for per-penalty `epsilon` and the gap source; the policy metadata records objective bounds and fallbacks but not per-penalty `epsilon`/source or the margin. Either thread `_gap_source`/`epsilon` into the `context` metadata or drop the binding.

- Test gaps vs the issue's explicit unit-test list: no test exercises (a) a generated penalty with `epsilon != 1` (the `PBO.mingap` branch of `_positive_penalty_gap`; all current tests hit the integer-valued `epsilon = 1` shortcut), or (b) any fallback path (`fallback_count > 0`, a populated `fallbacks` entry). Both are new logic. Recommend adding targeted unit tests.

## 3. Questions

- "Unbounded objective fallback" acceptance criterion. The range derives from the already-encoded `model.f`, whose `PBO.bounds` are always finite, so the `!objective_bounds["finite"]` branch ([src/compiler/penalties.jl:13](src/compiler/penalties.jl)) is effectively unreachable and the issue's required "unbounded objective fallback" unit test is absent. This is not a correctness bug (a finite bound is always valid), but the criterion is not met as written. Is the intent that the compiled-PBF approach makes finite certification always succeed (so the fallback is reserved for degenerate `epsilon`)? If so, document that; otherwise the unbounded path needs to be reachable and tested.

- `Closes #172` vs partial coverage. The core ask — objective-range exact-penalty as the certifiable default, `LegacyPenalty` as explicit policy/fallback, preserved hints, metadata surface, tightened network/routing diagnostics, docs, and the main-path tests — is delivered, so auto-closing is defensible. But several listed acceptance items are partial (unbounded-fallback test, `epsilon != 1` test, per-penalty `epsilon`/margin metadata, and the requested docs example showing network05 `1000000`/`1000001`, which lives in the benchmark report rather than the manual). If the team treats those as mandatory, consider `Refs #172` plus follow-ups; otherwise note the deferrals on the issue before it auto-closes.

- `benchmarks/qoblib/network_pilot.jl:60` repoints `penalty_scaling_follow_up` to #172 — the issue this PR closes. After merge it will reference a closed issue as a "follow-up." Point it at an open tracker (or drop it), as the routing pilot does by keeping #162.

## 4. Tests run and outcomes

All on a clean checkout of the PR head (`f499f19`) against merge-base `b1406fe`; diff confirmed at 15 files / +471 / -125, matching the PR totals.

- Full suite: `julia --project=. -e 'using Pkg; Pkg.test()'` — `♡ ToQUBO.jl 0.4.1 Test Suite ♡ | 1500 1500` pass, `Testing ToQUBO tests passed` (exit 0). `Penalty inference` testset 28/28. The QOBLib network golden test now asserts `objective_range == 1000000`, `largest_applied_penalty == 1000001`, and `constraint_penalties == [1000001.0]`, validating the issue's central claim.
- Standalone probe (outside the test env, since `QUBODrivers` only resolves under `Pkg.test`): confirmed the default policy is `ObjectiveRangePenalty`, metadata bounds populate correctly, and `PBO.bounds(model.f)` over-approximates the range for one-hot encodings (basis for the nonblocking tightness note above).
- Did not rebuild the docs (`docs/make.jl`) myself; the four new attributes added to the `@docs` block all carry docstrings, so Documenter should resolve them, and the PR reports a successful `--skip-deploy` build. Flagging that I did not independently run it.

Spot-checks beyond the tests: verified the sign/scale/offset arithmetic (`MAX_SENSE`, range 2, `epsilon` 1 → `-3.0`) matches both the test and the formula; confirmed `LinearPenalty` still errors without an explicit hint (unchanged); confirmed the new default is `>=` the legacy penalty whenever `epsilon = 1` and `range >= delta`, so no exactness regression.

## 5. Merge readiness

Mergeable on the merits: the math is sound, the default tightens penalties to the certified exact-penalty bound, existing behavior (hints, `LegacyPenalty`, scale/offset) is preserved, and the full suite passes. The open items are nonblocking quality/coverage gaps (fallback and `epsilon != 1` tests, unused `_gap_source`, the one-hot range caveat, and the #172 self-referential follow-up link). I would be comfortable with this merging once another maintainer records the formal approval, ideally after the fallback/`epsilon != 1` tests are added or the unbounded-fallback criterion is explicitly deferred on #172.

bernalde:
Addressed review comments in `803f1b2b9d3ffacd0c6eb4a406e5a833a6a9ee13`.

Commits pushed:
- `803f1b2` Address penalty policy review comments

Main changes:
- Qualified the exact-penalty documentation with the `PenaltyScale()` condition.
- Documented that compiled-PBF interval bounds are sufficient and finite for normal finite encoded models, but can overestimate the exact source objective span.
- Added `inferred_penalties` metadata with per-inferred penalty `epsilon`, `epsilon_source`, selected policy, scale/offset, and final coefficient.
- Added tests for the `PBO.mingap` path with `epsilon = 0.5` and for recorded fallback metadata.
- Removed the network pilot's `penalty_scaling_follow_up` link to #172 and updated the golden report/tests.

Tests run:
- `julia --project=. -e 'using Test; import MathOptInterface as MOI; import ToQUBO; using ToQUBO: Attributes, Encoding; include("test/unit/compiler/penalties.jl"); test_compiler_penalties()'`: passed, 55/55.
- `julia --project=. -e 'import Pkg; Pkg.test()'`: passed, 1525/1525.
- `julia --project=docs docs/make.jl --skip-deploy`: passed.
- `git diff --check`: passed.

- None.

Remaining risks or follow-up:
- No separate project type-check command was found. The compiled-PBF range remains a sufficient interval bound rather than the tightest source-domain range; this is now documented.

docs/src/manual/4-settings.md
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/803f1b2b9d3ffacd0c6eb4a406e5a833a6a9ee13. The docs now state that `β > 0` is sufficient at the default `s = 1`, and that custom scales require `s * (U - L + β) > U - L` for the same exactness guarantee.

src/compiler/penalties.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/803f1b2b9d3ffacd0c6eb4a406e5a833a6a9ee13. The docs now explain that compiled-PBF interval bounds are finite for normal finite encoded models and may overestimate the exact source span under encodings such as one-hot, which keeps the bound sufficient but not always tight.

src/compiler/penalties.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/803f1b2b9d3ffacd0c6eb4a406e5a833a6a9ee13. `_gap_source` is now recorded in `inferred_penalties` metadata as `epsilon_source`, together with `epsilon`, selected policy, scale/offset, and final penalty. `test/unit/compiler/penalties.jl` now covers `epsilon = 0.5` through `PBO.mingap` and a fallback metadata path.

benchmarks/qoblib/network_pilot.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/803f1b2b9d3ffacd0c6eb4a406e5a833a6a9ee13. Removed `penalty_scaling_follow_up` from the network pilot provenance and updated the golden report/tests, since #172 is the issue this PR is intended to close.

src/attributes/compiler.jl
```diff
@@ -241,16 +280,75 @@ function ignore_feasible_constraints(model::Optimizer)::Bool
...
+end
+
+function penalty_policy(model::Optimizer)::AutomaticPenaltyPolicy
+    return MOI.get(model, PenaltyPolicy())
+end
+
+@doc raw"""
+    PenaltyPolicyMetadata()
+
+Return diagnostic metadata for the automatic penalty policy used during the
```
bernalde, line 318: Nonblocking (round 2, new metadata schema): this docstring now understates the returned contract. `PenaltyPolicyMetadata` also returns `inferred_penalties` (per-penalty `epsilon`, `epsilon_source`, `scale`, `offset`, `automatic_policy`, `selected_policy`, optional `fallback_reason`) and `fallback_count`. Since this is a public attribute and the payload is advertised in NEWS.md, document the `inferred_penalties` structure here, and note it covers only inferred penalties (explicit hints bypass `_infer_and_record_penalty_factor` and are not recorded).

src/compiler/penalties.jl
```diff
@@ -3,15 +3,201 @@ function _uses_linear_equality_penalty(model::Virtual.Model, ci::CI)::Bool
...
+
+function _inferred_penalty_bucket(kind::String)
+    if kind == "constraint"
+        return "constraints"
+    elseif kind == "variable"
+        return "variables"
+    elseif kind == "slack_variable"
+        return "slack_variables"
+    else
+        return kind
```
bernalde: Nonblocking (round 2): the `else return kind` fallthrough is misleading. `context["inferred_penalties"]` only contains the three pre-created buckets (`constraints`/`variables`/`slack_variables`), so an unrecognized `kind` would `KeyError` on the push at line 159 rather than degrade gracefully. All three call sites pass known kinds, so this is unreachable, but the branch implies handling that does not exist. Drop the `else` (error loudly) or `@assert kind in ("constraint","variable","slack_variable")`.

test/unit/compiler/penalties.jl
```diff
@@ -69,20 +73,119 @@ function _variable_penalty_test_model(; scale = nothing, offset = nothing)
...
+    @test penalty == -3.0
+    @test metadata["policy"] == "ObjectiveRangePenalty"
+    @test metadata["objective_bounds"]["source"] == "compiled_pbf_bounds"
+    @test metadata["objective_bounds"]["lower"] == 0.0
+    @test metadata["objective_bounds"]["upper"] == 2.0
+    @test metadata["objective_bounds"]["range"] == 2.0
+    @test metadata["objective_bounds"]["finite"] === true
+    @test metadata["fallback_count"] == 0
+    @test isempty(metadata["fallbacks"])
+    inferred = only(metadata["inferred_penalties"]["constraints"])
```
bernalde, line 128: Nonblocking (round 2): the new `inferred_penalties` assertions only cover the `constraints` bucket here (and in the fractional/fallback tests). The `variables` and `slack_variables` buckets are populated but never asserted. For a new public metadata surface, add at least one assertion on a variable-encoding or slack inferred entry — the slack-encoding fixture in `test_compiler_applied_penalty_metadata` could carry it.

bernalde (commented):
Follow-up review (round 2), scoped to the new `inferred_penalties` penalty-policy metadata schema introduced in `803f1b2`. The four prior review threads were verified as addressed and are now resolved. This is my own PR, so GitHub only permits a COMMENT event from me; the formal verdict must come from another maintainer.

I re-ran the full suite on `803f1b2`: 1525/1525 pass (`Penalty inference` 55/55), exit 0. The new schema is correct and self-consistent: the fallback-association logic in `_infer_and_record_penalty_factor` correctly attributes a just-pushed fallback to the current penalty, `automatic_policy` vs `selected_policy` is a sound distinction (explicit `LegacyPenalty` is not recorded as a fallback), and the recorded `penalty` equals the applied coefficient. No blocking issues.

## 1. Blocking issues

None.

## 2. Nonblocking issues

- Public docstring is now stale relative to the returned contract. `PenaltyPolicyMetadata` ([src/attributes/compiler.jl:318](src/attributes/compiler.jl)) documents only "objective bounds and any fallback reasons," but the returned dict now also carries `inferred_penalties` (per-penalty `epsilon`, `epsilon_source`, `scale`, `offset`, `automatic_policy`, `selected_policy`, optional `fallback_reason`) and `fallback_count`. Since this is a public attribute and the payload is advertised in NEWS.md, the docstring should describe the `inferred_penalties` structure and note it covers only inferred penalties (explicit hints are excluded, since hinted coefficients bypass `_infer_and_record_penalty_factor`).

- `_inferred_penalty_bucket` dead/ misleading `else` branch. [src/compiler/penalties.jl:127](src/compiler/penalties.jl) returns `kind` for an unknown `kind`, but `context["inferred_penalties"]` only contains the three pre-created buckets, so a push with an unrecognized bucket name would `KeyError` rather than degrade. All three call sites pass known kinds so it is unreachable, but the fallthrough implies graceful handling that does not exist. Drop the `else` (let an unexpected kind error loudly) or `@assert kind in (...)`.

- New metadata buckets only partially tested. The three penalty tests assert `inferred_penalties["constraints"]` only; the `variables` and `slack_variables` buckets are populated but never asserted. For a new public surface, add at least one assertion on a variable-encoding or slack inferred entry (the existing slack-encoding fixture in `test_compiler_applied_penalty_metadata` could cover it).

## 3. Questions

- Scope/overlap with `applied_penalties`. `inferred_penalties` repeats the final `penalty` per item already present in `applied_penalties`, adding the inference diagnostics (`epsilon`, policy). That is reasonable, but confirm the intent is a diagnostics-only view restricted to inferred (non-hinted) penalties; if a consumer ever needs a complete per-item penalty list, `applied_penalties` remains the source of truth. A one-line note in the docstring (see above) would settle this.

## 4. Tests run and outcomes

Clean checkout at head `803f1b2` (merge-base `b1406fe`). Full suite: `julia --project=. -e 'using Pkg; Pkg.test()'` — `♡ ToQUBO.jl 0.4.1 Test Suite ♡ | 1525 1525` pass, exit 0. `Penalty inference` testset 55/55, including the two new tests exercising the `pbo_mingap` epsilon path (`epsilon == 0.5`, `epsilon_source == "pbo_mingap"`) and the reachable fallback path via `_UnknownPenaltyPolicy` (`fallback_count == 1`, `selected_policy == "LegacyPenalty"`). I re-derived the fallback-attribution logic by reading rather than only trusting the tests; it is correct.

## 5. Merge readiness

Mergeable. The new metadata schema is correct, consistent with the existing `applied_penalties` bucketing convention, and covered for the primary path; the remaining items are documentation and test-coverage polish, all nonblocking. The four original review points are resolved. I would be comfortable with this merging once another maintainer records the formal approval.

bernalde:
Commits pushed:
- 036c02f94589f9881a2b5a0f0ba8a19cbfee6b0f Polish penalty metadata diagnostics

Main changes made:
- Expanded the `PenaltyPolicyMetadata` documentation to describe `inferred_penalties`, its buckets and fields, and the fact that explicit hints are excluded.
- Changed `_inferred_penalty_bucket` to fail explicitly on unknown kinds instead of returning a misleading fallback bucket.
- Added unit assertions for `inferred_penalties["variables"]` and `inferred_penalties["slack_variables"]`.

Tests run and results:
- `julia --project=. -e 'using Test; import MathOptInterface as MOI; import ToQUBO; using ToQUBO: Attributes, Encoding; include("test/unit/compiler/penalties.jl"); test_compiler_penalties()'` passed, 67/67.\n- `julia --project=. -e 'import Pkg; Pkg.test()'` passed, 1537/1537.\n- `julia --project=docs docs/make.jl --skip-deploy` passed.\n- `git diff --check` passed.\n\nComments intentionally not addressed:\n- None.\n\nRemaining risks or follow-up items:\n- No separate type-check command was found beyond the Julia test suite and docs build.

src/attributes/compiler.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/036c02f94589f9881a2b5a0f0ba8a19cbfee6b0f. Expanded the `PenaltyPolicyMetadata` docstring to describe `inferred_penalties`, its buckets and fields, and that explicit hints are excluded; it points users to `AppliedPenalty`/`applied_penalties` for complete applied coefficients.

src/compiler/penalties.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/036c02f94589f9881a2b5a0f0ba8a19cbfee6b0f. `_inferred_penalty_bucket` now errors explicitly on an unknown kind instead of returning a misleading bucket name.

test/unit/compiler/penalties.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/036c02f94589f9881a2b5a0f0ba8a19cbfee6b0f. Added `inferred_penalties["variables"]` assertions in the variable penalty test and `inferred_penalties["slack_variables"]` assertions in `test_compiler_applied_penalty_metadata`.

PR #174, opened 2026-06-22
Resolve routing penalty-scaling follow-up

### LINKED ISSUE #162: Track QOBLib routing penalty-scaling divergence

bernalde:
## Summary

- Record that the issue-162 QOBLib routing penalty-scaling divergence no longer reproduces under the default objective-range penalty policy.
- Add QOBLib/Qiskit converter evidence for the routing pilot, including the converter pipeline and default-penalty interpretation.
- Add exact target-variable accounting showing that QOBLib/Qiskit retains 22 redundant routing bounds as 176 slack bits that ToQUBO drops.
- Regenerate the routing pilot report and update benchmark assertions for the resolved scaling conclusion.

## Resolution

QOBLib converts the routing LP through `misc/convert_lp2qubo.py`: Gurobi parses the LP, Qiskit `QuadraticProgramToQubo()` converts it with its default penalty, and QOBLib writes a symmetrized QS matrix. For this integer-coefficient routing pilot, Qiskit's default equality penalty is the same objective-range value ToQUBO now uses: `16697.376350318606`.

The remaining differences are not evidence that routing-specific penalty scaling is needed. The target-size delta is exact: both converters encode the 420 arc variables plus 21 load variables into 588 source bits, but QOBLib/Qiskit keeps 21 redundant `y[i] <= 231` bounds plus the redundant depot lower bound `y[1] >= 0`. Each retained `0..231` slack contributes eight binary variables, explaining QOBLib's 176 additional target variables.

Coefficient-wise, ToQUBO does not exactly reproduce the pinned QS metrics row, but the largest absolute QOBLIB-style coefficient is slightly smaller than QOBLib's pinned largest absolute coefficient. ToQUBO's positive-side maximum is larger, so exact coefficient extrema should be interpreted as conversion-convention artifacts rather than a penalty-scaling defect.

## Tests

- `julia --project=. benchmarks/qoblib/routing_pilot.jl`
- `julia --project=. -e 'using Test; include("test/unit/qoblib_benchmark.jl"); test_qoblib_benchmark_pilot()'` (388/388)
- `git diff --check`
- `julia --project=. -e 'import Pkg; Pkg.test()'` (1567/1567)

## Branch Hygiene

- Base branch: `main`
- Source branch point: `origin/main` at `713cec4b89714a5f57f2b1cf66908fc7527da7b0`
- Stacked status: not stacked
- Prerequisite PRs: none

Closes #162.

### DIFF
--- a/benchmarks/qoblib/routing_pilot.jl
+++ b/benchmarks/qoblib/routing_pilot.jl
@@ -154,6 +154,17 @@ const QOBLIB_QUBO_METRICS = Dict{String,Any}(
     "max_coeff" => 7_397_605_618.245156,
 )
 
+const LOAD_BINARY_BITS = ceil(Int, log2(CAPACITY + 1))
+const QOBLIB_REDUNDANT_CAPACITY_UPPER_BOUND_CONSTRAINTS = NUM_NODES
+const QOBLIB_REDUNDANT_DEPOT_LOWER_BOUND_CONSTRAINTS = 1
+const QOBLIB_REDUNDANT_SLACK_BITS =
+    LOAD_BINARY_BITS * (
+        QOBLIB_REDUNDANT_CAPACITY_UPPER_BOUND_CONSTRAINTS +
+        QOBLIB_REDUNDANT_DEPOT_LOWER_BOUND_CONSTRAINTS
+    )
+const COEFFICIENT_SAME_ORDER_LOWER = 0.1
+const COEFFICIENT_SAME_ORDER_UPPER = 10.0
+
 const QOBLIB_CONVERTER_EVIDENCE = Dict{String,Any}(
     "manual_verification" => true,
     "converter_path" => "misc/convert_lp2qubo.py",
@@ -165,22 +176,30 @@ const QOBLIB_CONVERTER_EVIDENCE = Dict{String,Any}(
         "QOBLIB reads the LP with Gurobi, changes any continuous variables to integer, builds a Qiskit QuadraticProgram with from_gurobipy, converts it with QuadraticProgramToQubo() using the converter default penalty, writes linear coefficients on the diagonal, symmetrizes Q as (Q + Q') / 2, and writes the objective offset separately.",
     "qiskit_converter_pipeline" =>
         "QuadraticProgramToQubo first handles a narrow set of special binary inequalities, then converts remaining inequalities to equalities with integer slack variables, encodes integer variables to binary, and finally applies equality penalties.",
+    "qiskit_optimization_version_checked" => "0.7.0",
+    "qiskit_penalty_formula_checked" =>
+        "LinearEqualityToPenalty._auto_define_penalty and LinearInequalityToPenalty._auto_define_penalty return 1 plus the objective linear/quadratic coefficient bound range for integer-coefficient constraints.",
     "qiskit_default_penalty_note" =>
-        "For integer-coefficient constraints, Qiskit's automatic equality penalty is 1 plus the objective coefficient bound range. For this routing pilot that matches ToQUBO's objective-range penalty 16697.376350318606.",
+        "For integer-coefficient constraints, Qiskit's automatic equality penalty is 1 plus the objective coefficient bound range. The pinned QOBLIB converter uses that default path, and for this routing pilot the formula matches ToQUBO's objective-range penalty 16697.376350318606.",
 )
 
 const ROUTING_CONVERTER_ACCOUNTING = Dict{String,Any}(
     "source_arc_binary_variables" => length(ARCS),
     "source_load_variables" => NUM_NODES,
-    "source_load_binary_variables" => 8 * NUM_NODES,
-    "source_binary_variables_after_encoding" => length(ARCS) + 8 * NUM_NODES,
-    "toqubo_redundant_constraints_dropped" => 22,
-    "qoblib_redundant_capacity_upper_bound_constraints" => NUM_NODES,
-    "qoblib_redundant_depot_lower_bound_constraints" => 1,
-    "qoblib_redundant_slack_bits" => 176,
+    "source_load_binary_variables" => LOAD_BINARY_BITS * NUM_NODES,
+    "source_binary_variables_after_encoding" =>
+        length(ARCS) + LOAD_BINARY_BITS * NUM_NODES,
+    "toqubo_redundant_constraints_dropped" =>
+        QOBLIB_REDUNDANT_CAPACITY_UPPER_BOUND_CONSTRAINTS +
+        QOBLIB_REDUNDANT_DEPOT_LOWER_BOUND_CONSTRAINTS,
+    "qoblib_redundant_capacity_upper_bound_constraints" =>
+        QOBLIB_REDUNDANT_CAPACITY_UPPER_BOUND_CONSTRAINTS,
+    "qoblib_redundant_depot_lower_bound_constraints" =>
+        QOBLIB_REDUNDANT_DEPOT_LOWER_BOUND_CONSTRAINTS,
+    "qoblib_redundant_slack_bits" => QOBLIB_REDUNDANT_SLACK_BITS,
     "target_variable_delta_explained_by_redundant_slack_bits" => true,
     "target_variable_accounting_note" =>
-        "Both converters use eight binary variables for each 0..231 load variable. QOBLIB/Qiskit retains 21 redundant y[i] <= 231 constraints and the redundant depot lower-bound constraint y[1] >= 0; each retained 0..231 slack contributes eight binary variables, explaining the 176-variable target delta exactly.",
+        "Both converters use $(LOAD_BINARY_BITS) binary variables for each 0..231 load variable. QOBLIB/Qiskit retains $(QOBLIB_REDUNDANT_CAPACITY_UPPER_BOUND_CONSTRAINTS) redundant y[i] <= 231 constraints and the redundant depot lower-bound constraint y[1] >= 0; each retained 0..231 slack contributes $(LOAD_BINARY_BITS) binary variables, explaining the $(QOBLIB_REDUNDANT_SLACK_BITS)-variable target delta exactly.",
 )
 
 const KNOWN_INCUMBENT = Dict{String,Any}(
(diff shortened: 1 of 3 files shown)


bernalde (commented):
Blocking: none.

Nonblocking: none.

Question: none.

Issue fit: The PR addresses #162's core ask. It verifies the post-#173 routing coefficient behavior, documents QOBLib/Qiskit's converter path, explains why exact coefficient parity is not expected, and accounts for the full 176-variable target-size delta through redundant retained slacks. `Closes #162` is appropriate.

Tests run and outcomes:
- `julia --project=. benchmarks/qoblib/routing_pilot.jl` passed.
- `julia --project=. -e 'using Test; include("test/unit/qoblib_benchmark.jl"); test_qoblib_benchmark_pilot()'` passed, 388/388.
- `git diff --check` passed.
- `julia --project=. -e 'import Pkg; Pkg.test()'` passed, 1567/1567.
- `julia --project=docs docs/make.jl --skip-deploy` passed.

Merge-readiness: I would consider this merge-ready once GitHub CI finishes green and another maintainer provides the formal approval. This review is submitted as COMMENT because GitHub rejects APPROVE/REQUEST_CHANGES on my own PR.

benchmarks/qoblib/routing_pilot.jl
```diff
@@ -154,6 +154,35 @@ const QOBLIB_QUBO_METRICS = Dict{String,Any}(
...
+    "converter_path" => "misc/convert_lp2qubo.py",
+    "converter_line_refs" => [
+        "misc/convert_lp2qubo.py:55-65",
+        "misc/convert_lp2qubo.py:82-89",
+    ],
+    "converter_convention" =>
+        "QOBLIB reads the LP with Gurobi, changes any continuous variables to integer, builds a Qiskit QuadraticProgram with from_gurobipy, converts it with QuadraticProgramToQubo() using the converter default penalty, writes linear coefficients on the diagonal, symmetrizes Q as (Q + Q') / 2, and writes the objective offset separately.",
+    "qiskit_converter_pipeline" =>
+        "QuadraticProgramToQubo first handles a narrow set of special binary inequalities, then converts remaining inequalities to equalities with integer slack variables, encodes integer variables to binary, and finally applies equality penalties.",
+    "qiskit_default_penalty_note" =>
```
bernalde, line 182: Question: this asserts Qiskit's default equality penalty equals ToQUBO's objective-range penalty 16697.376350318606. The ToQUBO side is genuinely verified (the test checks `only(metadata["constraint_penalties"]) ≈ 16697.376350318606` against compiler output), but the Qiskit value and the converter pipeline description rest on `manual_verification => true` and are only checked via `occursin` on this note string, so CI cannot catch drift. Acceptable per the network-pilot convention, but please confirm the converter line refs and the penalty derivation were checked against pinned commit `a686aaa`.

bernalde (commented):

This is a diagnostics-only change: it touches the routing benchmark pilot, its generated report, and the QOBLib benchmark test. No production compiler code is modified, so there is no behavioral regression surface. I verified the behavior by running it: full suite green (1567/1567), the routing pilot regenerates byte-identically to the checked-in report, and the central numeric claims are asserted against ToQUBO's actual compiler output.

## 1. Blocking issues

None.

## 2. Nonblocking issues

- Editorial conclusions are stored as literals and self-asserted. `penalty_scaling_divergence_resolved => true` and `routing_specific_scaling_needed => false` ([benchmarks/qoblib/routing_pilot.jl:587-588](benchmarks/qoblib/routing_pilot.jl)) are hardcoded booleans, and the test asserts them `=== true`/`=== false`, which cannot fail by construction. The substantive guard against a future penalty regression is the computed coefficient-ratio assertions (`qoblib_symmetric_abs_bound_ratio` ≈ 0.9657, etc.), which are tied to real output — so the conclusion is backed, but the booleans themselves do not self-validate. Consider deriving the "resolved" flag from a threshold on the computed ratio so the conclusion tracks the metric, or at least add a comment noting the ratio assertions are the real guard.

- `qoblib_redundant_slack_bits => 176` ([benchmarks/qoblib/routing_pilot.jl:180](benchmarks/qoblib/routing_pilot.jl)) is a magic literal. It equals `8 * (qoblib_redundant_capacity_upper_bound_constraints + qoblib_redundant_depot_lower_bound_constraints)` = `8 * 22`. Deriving it (and the `-176` delta consistency check already does compare against it) would keep the accounting self-consistent if the instance constants ever change. Low severity given this is a single pinned-instance pilot, consistent with the other literals here.

## 3. Questions

- The Qiskit-equivalence claim is manual, not reproduced in CI. The report states Qiskit's default equality penalty equals ToQUBO's objective-range penalty `16697.376350318606` ([benchmarks/qoblib/routing_pilot.jl:168](benchmarks/qoblib/routing_pilot.jl)). The ToQUBO side is genuinely verified — the test asserts `only(metadata["constraint_penalties"]) ≈ 16697.376350318606` against the compiler output (test/unit/qoblib_benchmark.jl). The Qiskit side and the "22 redundant bounds → 176 slack bits" breakdown rest on `manual_verification => true` and are only checked via `occursin` on the note strings, so CI cannot catch drift if the upstream converter behavior is mischaracterized. This matches the network pilot's convention for analogous Qiskit claims, so it is acceptable, but confirm the converter line refs (`misc/convert_lp2qubo.py:55-65`, `:82-89`) and the slack-range reasoning were checked against the pinned QOBLIB commit `a686aaa`.

- `Closes #162` is appropriate, not a `Refs` case. Issue #162 is a tracking/investigation issue asking whether the divergence is expected, whether routing-specific scaling is needed, and how QOBLib treats the MTZ load variables / capacity bounds / redundant constraints. The PR answers all three: the ~3-orders-of-magnitude divergence no longer reproduces under the merged objective-range policy (abs-bound ratio now ≈ 0.97), no routing-specific scaling is needed, and the redundant-bound/slack accounting is documented. Closing the investigation issue on merge is correct. Note the resolution depends on PR #173, which is already in the merge-base.

- Minor naming drift: the report key is still `report["follow_up"]` but now renders under a `## Penalty Scaling Resolution` heading, and the provenance field remains `penalty_scaling_follow_up` while the report relabels it "Penalty scaling issue". Cosmetic; the #162 link now reads as a back-reference to the resolved investigation, which is fine.

## 4. Tests run and outcomes

All on a clean checkout of the PR head (`2d439d3`) against merge-base `713cec4` (which is the PR #173 merge commit, so the objective-range policy this pilot relies on is present); diff confirmed at 3 files / +218 / -10, matching the PR totals.

- Full suite: `julia --project=. -e 'using Pkg; Pkg.test()'` — `♡ ToQUBO.jl 0.4.1 Test Suite ♡ | 1567 1567` pass, `Testing ToQUBO tests passed` (exit 0). `QOBLib benchmark pilot` testset 388/388. Matches the PR's reported counts; I re-ran rather than trusting them.
- Report reproducibility: `julia --project=. benchmarks/qoblib/routing_pilot.jl <out>` regenerates output byte-identical (after CRLF normalization) to the checked-in `benchmarks/qoblib/reports/routing_pilot.md`.
- Spot-checked the live-computed assertions: `target_variable_delta_vs_qoblib_qs == -176` with `target_variable_delta_explained_by_redundant_slack_bits === true` (delta computed from `target["num_variables"]`, not hardcoded), and the applied penalty `≈ 16697.376350318606` from `metadata["constraint_penalties"]`.

## 5. Merge readiness

Mergeable. The change is documentation/diagnostics only, deterministic, faithfully closes the #162 investigation, and the full suite passes. The open items are nonblocking quality notes (self-asserting conclusion booleans, a magic literal) and a question about CI coverage of the manual Qiskit claim. I would be comfortable with this merging once another maintainer records the formal approval.

benchmarks/qoblib/routing_pilot.jl
```diff
@@ -154,6 +154,35 @@ const QOBLIB_QUBO_METRICS = Dict{String,Any}(
...
+
+const ROUTING_CONVERTER_ACCOUNTING = Dict{String,Any}(
+    "source_arc_binary_variables" => length(ARCS),
+    "source_load_variables" => NUM_NODES,
+    "source_load_binary_variables" => 8 * NUM_NODES,
+    "source_binary_variables_after_encoding" => length(ARCS) + 8 * NUM_NODES,
+    "toqubo_redundant_constraints_dropped" => 22,
+    "qoblib_redundant_capacity_upper_bound_constraints" => NUM_NODES,
+    "qoblib_redundant_depot_lower_bound_constraints" => 1,
+    "qoblib_redundant_slack_bits" => 176,
```
bernalde: Nonblocking (maintainability): `qoblib_redundant_slack_bits => 176` is a magic literal equal to `8 * (qoblib_redundant_capacity_upper_bound_constraints + qoblib_redundant_depot_lower_bound_constraints)` = `8 * 22`. Deriving it would keep the accounting self-consistent if the instance constants change. Low severity for a single pinned-instance pilot.

benchmarks/qoblib/routing_pilot.jl
```diff
@@ -541,6 +575,19 @@ function _comparison(target, metadata)
...
+        "qoblib_symmetric_min_to_canonical_min_abs_ratio" =>
+            abs(target["qoblib_symmetric_min_coeff"]) /
+            abs(QOBLIB_QUBO_METRICS["min_coeff"]),
+        "qoblib_symmetric_max_to_canonical_max_abs_ratio" =>
+            abs(target["qoblib_symmetric_max_coeff"]) /
+            abs(QOBLIB_QUBO_METRICS["max_coeff"]),
+        "qoblib_symmetric_abs_bound_ratio" =>
+            max(abs(target["qoblib_symmetric_min_coeff"]), abs(target["qoblib_symmetric_max_coeff"])) /
+            max(abs(QOBLIB_QUBO_METRICS["min_coeff"]), abs(QOBLIB_QUBO_METRICS["max_coeff"])),
+        "penalty_scaling_divergence_resolved" => true,
```
bernalde: Nonblocking: `penalty_scaling_divergence_resolved => true` and `routing_specific_scaling_needed => false` are hardcoded conclusions, and the test asserts them `=== true`/`=== false`, which cannot fail by construction. The real regression guard is the computed `qoblib_symmetric_abs_bound_ratio` (≈ 0.9657) assertion. Consider deriving the `resolved` flag from a threshold on that ratio so the conclusion tracks the metric, or add a comment noting the ratio assertions are the actual guard.

bernalde:
Commit pushed:
- `63bd289` `fix: address routing pilot review feedback (#162)`

Main changes:
- Derived the QOBLib redundant slack-bit accounting from routing constants instead of storing `176` as a magic literal.
- Derived `penalty_scaling_divergence_resolved` and `routing_specific_scaling_needed` from the computed QOBLIB-style largest absolute coefficient ratio and explicit same-order bounds.
- Expanded the converter evidence to name the checked Qiskit optimization version and the auto-penalty formula used by `LinearEqualityToPenalty` / `LinearInequalityToPenalty`.
- Regenerated the routing pilot report and extended QOBLib benchmark assertions for the new fields.

Tests:
- `julia --project=. benchmarks/qoblib/routing_pilot.jl` passed.
- `julia --project=. -e 'using Test; include("test/unit/qoblib_benchmark.jl"); test_qoblib_benchmark_pilot()'` passed, 395/395.
- `git diff --check` passed.
- `julia --project=. -e 'import Pkg; Pkg.test()'` passed, 1574/1574.
- `julia --project=docs docs/make.jl --skip-deploy` passed.

- None.

Remaining risks or follow-up:
- The Qiskit conversion is still documented as manual verification, consistent with the existing QOBLib network pilot pattern; no Qiskit/Gurobi reproduction is added to CI.

benchmarks/qoblib/routing_pilot.jl
bernalde: Addressed in `63bd289`. I expanded the converter evidence to record `qiskit_optimization_version_checked = "0.7.0"` and the exact auto-penalty formula checked for `LinearEqualityToPenalty` / `LinearInequalityToPenalty`, while keeping the pinned QOBLIB converter refs in the report.

benchmarks/qoblib/routing_pilot.jl
bernalde: Addressed in `63bd289`. `qoblib_redundant_slack_bits` is now derived from `LOAD_BINARY_BITS` and the retained redundant-bound counts instead of being stored as the literal `176`; the report was regenerated and tests cover the accounting.

benchmarks/qoblib/routing_pilot.jl
bernalde: Addressed in `63bd289`. The resolved/scaling-needed flags now derive from the computed `qoblib_symmetric_abs_bound_ratio` and explicit same-order ratio bounds, and the report/test output exposes those bounds.

PR #185, opened 2026-06-23
Optimize dense QUBO fast path

### LINKED ISSUE #183: Optimize QUBO fast path for dense squared-affine objectives
## Context

The refreshed `ToQUBO-benchmark` run with the released QUBO stack shows that ToQUBO is now very competitive on TSP but still slower than Amplify on NPP-like dense squared-affine objectives.

Benchmark source/result:

- Benchmark repository: `JuliaQUBO/ToQUBO-benchmark`
- Benchmark commit with CI-generated results/plots: `2bd47157de532fa143f63d8d353ee6a3338c75ef`
- Benchmark workflow: https://github.com/JuliaQUBO/ToQUBO-benchmark/actions/runs/28054304889
- Plot workflow: https://github.com/JuliaQUBO/ToQUBO-benchmark/actions/runs/28056410120
- Result set id: `released-qubo-stack-2026-06-23`

Resolved package versions in the benchmark report:

- Julia: `1.12.6`
- ToQUBO: `0.5.0`
- QUBOTools: `0.14.4`
- JuMP: `1.30.1`
- Amplify: `1.6.1`
- PyQUBO: `1.5.0`
- dimod: `0.12.22`
- qiskit-optimization: `0.7.0`
- qubovert: `1.2.5`

## Benchmark observation

For NPP at 1,000 variables, the latest benchmark reports:

```text
Amplify: 0.106837511 s
ToQUBO:  0.397157106 s at n=1000
ToQUBO:  0.450459659 s max row near n=980
PyQUBO:  0.784842730 s
```

For TSP at 10,000 variables, ToQUBO is already fastest among the compared packages:

```text
ToQUBO:   1.973833604 s
Amplify:  5.068821192 s
PyQUBO:  12.611128330 s
qubovert: 103.741641998 s
```

So the remaining gap is not a general ToQUBO slowdown. It appears specifically in NPP-like objectives where the model is essentially one dense squared affine expression:

```julia
@objective(model, Min, ((2x .- 1)'s)^2)
```

Equivalent form:

```math
(2 \sum_i s_i x_i - S)^2, \quad S = \sum_i s_i
```

For binary variables, this expands to:

```text
linear_i    = 4*s_i^2 - 4*S*s_i
quadratic_ij = 8*s_i*s_j, i < j
constant    = S^2
```

## Local profiling evidence

Using the benchmark project at `ToQUBO-benchmark` commit `2bd4715`, with the released environment above, a warmed local run at `n=1000` produced this phase split over 10 runs:

```text
model_time         median=0.181671 s  mean=0.188602 s
compiler_time      median=0.189907 s  mean=0.190901 s
convert_time       median=0.018137 s  mean=0.018373 s
total_time         median=0.387575 s  mean=0.397886 s
compiler_share     median=0.4858
model_share        median=0.4687
convert_share      median=0.0453
```

A more explicit split of the NPP JuMP construction and ToQUBO path at `n=1000` showed:

```text
t_model       median=0.001062 s
t_var         median=0.000182 s
t_expr        median=0.120597 s   # build dense quadratic expression
t_obj         median=0.112120 s   # set objective on JuMP/MOI model
t_compile     median=0.310858 s   # optimize!(model), noisy but consistently large
t_extract     median=0.010603 s   # QUBOTools.backend(model)
```

The Julia sampler points to the expected dense-objective hot spots:

- JuMP expression construction:
  - `JuMP/src/quad_expr.jl:add_to_expression!`
  - `JuMP/src/operators.jl:*`
  - `JuMP/src/macros.jl` / `MutableArithmetics`
  - `OrderedCollections.OrderedDict` insert/hash paths
- ToQUBO QUBO fast path:
  - `ToQUBO/src/compiler/compiler.jl:compile!`
  - `_copy_qubo_objective!`
  - `_copy_qubo_function!(::MOI.ScalarQuadraticFunction)`
  - `_add_qubo_quadratic_term!`
  - `_copy_qubo_hamiltonian!`
  - `_output_qubo_objective!`
  - `PseudoBooleanOptimization` dictionary copy/merge/hash paths
- Backend extraction:
  - `QUBOTools_MOI/model_parser.jl:_extract_bool_model`
  - `_append_quadratic_term!`
  - `_finalize_moi_model`

A direct coefficient aggregation baseline for NPP at `n=1000` is much smaller than the current full path:

```text
direct_arrays      median=0.000656 s
direct_sparse      median=0.007721 s
```

This direct baseline only generates the mathematically equivalent linear/quadratic coefficients and sparse storage. It is not a drop-in replacement for ToQUBO's full JuMP/MOI semantics, but it shows that the arithmetic and sparse finalization are not inherently expensive enough to explain the current 0.39-0.45 s NPP time.

## Current likely cause

The current QUBO path appears to materialize and copy the dense quadratic objective multiple times:

1. JuMP builds the dense `ScalarQuadraticFunction` for the squared affine objective.
2. ToQUBO detects the source model is already QUBO and enters `Compiler.copy!`.
3. `Compiler.copy!` copies the MOI quadratic objective into a `PseudoBooleanOptimization.PBF` dictionary.
4. It copies that PBF into `model.H`.
5. `_output_qubo_objective!` converts the PBF back into an MOI `ScalarQuadraticFunction` on `target_model`.
6. `QUBOTools.backend(model)` then parses the target MOI model back into a `QUBOTools.Model` sparse form.

For dense NPP-like objectives, those repeated dictionary and MOI conversions dominate. The backend extraction fix already made extraction small; the remaining gap is mostly compile/copy/materialization overhead.

## Proposed implementation direction

Add a faster QUBO fast path for unconstrained binary quadratic source models.

A likely first implementation is to optimize the existing `is_qubo(model.source_model)` branch in `Compiler.compile!` / `Compiler.copy!`:

1. Avoid copying `MOI.ScalarQuadraticFunction` terms into `PseudoBooleanOptimization.PBF` and then back into `MOI.ScalarQuadraticFunction` when no reformulation is needed.
2. Either copy the source quadratic function directly into `target_model` with only the required variable-index remapping, or build/cache the final `QUBOTools.Model` representation directly during compilation.
3. Let `QUBOTools.backend(model::ToQUBO.Optimizer)` return the cached direct backend when available instead of reparsing `target_model`.
4. Preserve the current public behavior of `optimize!(model)` and `QUBOTools.backend(model)`.
5. Keep the existing generic compilation path unchanged for constrained, higher-order, encoded, or penalty-generated models.

A more ambitious follow-up, probably separate from this issue, would be a dedicated squared-affine/front-end API that avoids JuMP's dense quadratic expression construction too. Without such a front-end change, the benchmark may still pay roughly 0.18-0.23 s in JuMP model construction even if ToQUBO compilation/extraction becomes near the direct sparse baseline. This issue should therefore focus first on removing avoidable ToQUBO-side overhead, not on promising that standard JuMP syntax will necessarily beat Amplify on NPP.

## Acceptance criteria

- Add a benchmark or regression test for an already-QUBO binary quadratic model with dense terms, ideally mirroring NPP at several sizes.
- Confirm that `optimize!(model)` still produces a valid QUBO target and that `QUBOTools.backend(model)` returns the same coefficients/sense/domain as the current path.
- Include correctness tests for:
  - diagonal quadratic terms, using binary `x^2 = x` handling;
  - off-diagonal term ordering;
  - affine terms;
  - constant term;
  - minimization/maximization sense;
  - variable map consistency;
  - empty/no-fixed-variable cases.
- Preserve fallback behavior for non-QUBO models.
- Track performance before/after on the benchmark NPP case.

Suggested performance target for this first pass:

- Reduce ToQUBO-side `compiler_time + convert_time` for NPP `n=1000` substantially from the current warmed median of roughly `0.208 s`.
- Ideally get ToQUBO-side compile/backend overhead close to the direct sparse baseline order of magnitude, while recognizing that standard JuMP expression construction is a separate cost.

## Reproduction notes

The profiling commands were run from `JuliaQUBO/ToQUBO-benchmark` at commit `2bd4715` with:

```bash
julia +release --project=benchmark/ToQUBO
```

The benchmark NPP implementation is in:

```text
benchmark/ToQUBO/problems.jl
```

Relevant current ToQUBO code paths from the registered package:

```text
src/compiler/compiler.jl
src/wrapper.jl
```

Relevant QUBOTools parser path observed by the sampler:

```text
ext/QUBOTools_MOI/model_parser.jl
```

bernalde:
## Summary

Closes #183.

This PR optimizes the already-QUBO path for unconstrained binary quadratic models. The compiler now remaps source objective terms directly into the target MOI QUBO model and builds a cached `QUBOTools.Model` backend in the same pass, avoiding the previous MOI -> PBO -> MOI copy and the later target-model reparse in `QUBOTools.backend`.

The generic reformulation path for constrained, encoded, penalty-generated, and non-binary models is unchanged.

## Tests

- `git diff --check`
- `julia --project -e 'using Pkg; Pkg.test()'`
  - Passed: 1639/1639 tests

`JuliaFormatter` was not run because it is not installed in the active project environment.

## Performance

Local MOI-level dense NPP fixture at `n = 1000`, with two warmup runs and ten measured runs:

- `origin/main`: median compile + backend extraction `0.4081747785 s`
- this branch: median compile + backend extraction `0.114548496 s`

This measures ToQUBO-side `MOI.optimize!(model)` plus `QUBOTools.backend(model)` for an already-QUBO dense quadratic objective.

## Branch Hygiene

- Base branch: `main`
- Source branch point: `origin/main` at `152bae86907fa2fe7da03c5e3a0cbccc0f8be801`
- Stacked status: not stacked
- Prerequisite PRs: none

### DIFF
--- a/src/compiler/compiler.jl
+++ b/src/compiler/compiler.jl
@@ -164,6 +164,12 @@ function _add_or_drop!(terms::Dict{K,T}, key::K, value::T) where {K,T}
     return nothing
 end
 
+function _add_qubo_linear!(terms::Dict{VI,T}, key::VI, value::T) where {T}
+    terms[key] = get(terms, key, zero(T)) + value
+
+    return nothing
+end
+
 function _qubo_backend_sense(sense::MOI.OptimizationSense)
     return sense === MOI.MIN_SENSE ? :min : :max
 end
@@ -240,7 +246,7 @@ function _copy_qubo_objective!(
         c = term.coefficient
 
         push!(affine_terms, SAT{T}(c, y))
-        _add_or_drop!(linear_terms, y, c)
+        _add_qubo_linear!(linear_terms, y, c)
     end
 
     MOI.set(
@@ -272,7 +278,7 @@ function _copy_qubo_objective!(
         push!(quadratic_terms, SQT{T}(c, x, y))
 
         if x == y
-            _add_or_drop!(backend_linear_terms, x, c / 2)
+            _add_qubo_linear!(backend_linear_terms, x, c / 2)
         else
             u, v = _ordered_qubo_pair(x, y)
             _add_or_drop!(backend_quadratic_terms, (u, v), c)
@@ -284,7 +290,7 @@ function _copy_qubo_objective!(
         c = term.coefficient
 
         push!(affine_terms, SAT{T}(c, y))
-        _add_or_drop!(backend_linear_terms, y, c)
+        _add_qubo_linear!(backend_linear_terms, y, c)
     end
 
     MOI.set(
--- a/src/compiler/compiler.jl
+++ b/src/compiler/compiler.jl
@@ -5,6 +5,7 @@ import MathOptInterface as MOI
 import MathOptInterface: empty!
 import PseudoBooleanOptimization as PBO
 
+import QUBOTools
 import QUBOTools: AbstractArchitecture, GenericArchitecture
 
 import ..Attributes
@@ -93,6 +94,7 @@ function reset!(model::Virtual.Model, ::AbstractArchitecture = GenericArchitectu
     Base.empty!(model.s)
     Base.empty!(model.η)
     Base.empty!(model.H)
+    model.qubo_backend_cache = nothing
 
     # Optimize-generated status
     MOI.set(model, Attributes.CompilationStatus(), nothing)
@@ -116,8 +118,6 @@ function Compiler.copy!(model::Virtual.Model{T}, arch::AbstractArchitecture) whe
     )
 
     _copy_qubo_objective!(model, variable_map)
-    _copy_qubo_hamiltonian!(model)
-    _output_qubo_objective!(model)
 
     model.compiler_settings[:qubo_fast_path] = true
 
@@ -148,82 +148,49 @@ function _copy_qubo_variables!(model::Virtual.Model{T}) where {T}
     return variable_map
 end
 
-function _qubo_term(x::VI)
-    return PBO.Term{VI}(VI[x])
+function _ordered_qubo_pair(x::VI, y::VI)
+    return x.value <= y.value ? (x, y) : (y, x)
 end
 
-function _qubo_term(x::VI, y::VI)
-    if x.value <= y.value
-        return PBO.Term{VI}(VI[x, y])
+function _add_or_drop!(terms::Dict{K,T}, key::K, value::T) where {K,T}
+    new_value = get(terms, key, zero(T)) + value
+
+    if iszero(new_value)
+        delete!(terms, key)
     else
-        return PBO.Term{VI}(VI[y, x])
-    end
-end
-
-function _add_qubo_affine_term!(f::PBO.PBF{VI,T}, variable_map, a::SAT{T}) where {T}
-    f[_qubo_term(variable_map[a.variable])] += a.coefficient
-
-    return nothing
-end
-
-function _add_qubo_quadratic_term!(f::PBO.PBF{VI,T}, variable_map, q::SQT{T}) where {T}
-    x = variable_map[q.variable_1]
-    y = variable_map[q.variable_2]
-    c = q.coefficient
-
-    if x == y
-        # MOI stores diagonal quadratic terms with doubled coefficients for
-        # pseudo-boolean variables because x^2 == x.
-        f[_qubo_term(x)] += c / 2
-    else
-        f[_qubo_term(x, y)] += c
+        terms[key] = new_value
     end
 
     return nothing
 end
 
-function _copy_qubo_function!(
-    f::PBO.PBF{VI,T},
-    variable_map,
-    vi::VI,
+function _qubo_backend_sense(sense::MOI.OptimizationSense)
+    return sense === MOI.MIN_SENSE ? :min : :max
+end
+
+function _empty_qubo_linear_terms(::Type{T}, variable_map::Dict{VI,VI}) where {T}
+    linear_terms = sizehint!(Dict{VI,T}(), length(variable_map))
+
+    for y in values(variable_map)
+        linear_terms[y] = zero(T)
+    end
+
+    return linear_terms
+end
+
+function _cache_qubo_backend!(
+    model::Virtual.Model{T},
+    linear_terms::Dict{VI,T},
+    quadratic_terms::Dict{Tuple{VI,VI},T},
+    offset::T,
 ) where {T}
-    f[_qubo_term(variable_map[vi])] += one(T)
-
-    return nothing
-end
-
-function _copy_qubo_function!(
-    f::PBO.PBF{VI,T},
-    variable_map,
-    obj::SAF{T},
-) where {T}
-    sizehint!(f, length(obj.terms) + 1)
-
-    for a in obj.terms
-        _add_qubo_affine_term!(f, variable_map, a)
-    end
-
-    f[nothing] += obj.constant
-
-    return nothing
-end
-
-function _copy_qubo_function!(
-    f::PBO.PBF{VI,T},
-    variable_map,
-    obj::SQF{T},
-) where {T}
-    sizehint!(f, length(obj.quadratic_terms) + length(obj.affine_terms) + 1)
-
-    for q in obj.quadratic_terms
-        _add_qubo_quadratic_term!(f, variable_map, q)
-    end
-
-    for a in obj.affine_terms
-        _add_qubo_affine_term!(f, variable_map, a)
-    end
-
-    f[nothing] += obj.constant
+    model.qubo_backend_cache = QUBOTools.Model{VI,T,Int}(
+        linear_terms,
+        quadratic_terms;
+        offset,
+        sense = _qubo_backend_sense(MOI.get(model.target_model, MOI.ObjectiveSense())),
+        domain = :bool,
+    )
 
     return nothing
 end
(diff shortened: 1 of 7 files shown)


bernalde (commented):
Senior maintainer review of #185 (Optimize dense QUBO fast path).

Note on numbering: the task referenced "PR 183", but #183 is the issue this PR closes; the pull request is #185. Reviewed at head 5c9a319 against merge-base 152bae8 (189 additions / 80 deletions across 6 files, matching the PR's reported totals). I am the PR author, so GitHub only allows a COMMENT review here; the formal APPROVE/REQUEST_CHANGES must come from another maintainer.

The change is sound in its core design: it remaps source objective terms straight into the target MOI model and builds a cached `QUBOTools.Model` in the same pass, dropping the old MOI -> PBO -> MOI round-trip and the later target reparse in `QUBOTools.backend`. The ordering question I worried about (cache uses `_ordered_qubo_pair`, the parse fallback keeps raw term order) is a non-issue: `QUBOTools._build_sparse_forms` normalizes both to the upper triangle and `sparse(...)` sums duplicate (row,col) entries, so the test's `x*y` + `y*x` pair correctly collapses to one coefficient in both paths.

Blocking issues:
- None.

Nonblocking issues:
1. Variable-set divergence between the cached backend and the parse fallback under explicit-zero or cancelling linear terms. See the inline comment on `_add_or_drop!` (src/compiler/compiler.jl). `_empty_qubo_linear_terms` pre-seeds every variable at zero, but `_add_or_drop!` then `delete!`s any linear key whose running sum hits zero. The fallback path `QUBOTools._extract_bool_model` (ext/QUBOTools_MOI/model_parser.jl:146) seeds `L = Dict(xi => 0 for xi in V)` and never deletes, so it keeps the variable. I reproduced this: an objective with an explicit `ScalarAffineTerm(0.0, z)` (or `+3z` and `-3z`) yields `QUBOTools.backend(model)` with 2 variables via the cache versus 3 via the parse path; the problem variable `z` is silently dropped from the cached QUBO. A binary variable absent from the objective entirely is unaffected (both report 3), so canonical JuMP/MOI input that prunes zeros is not hit. This is still a behavior regression versus main (which always used the parse path) and contradicts the issue's acceptance criterion that `QUBOTools.backend(model)` match the current path. Fix: match `_extract_bool_model` for linear terms (plain `terms[key] = get(terms, key, zero(T)) + value`, no delete); `_build_sparse_forms` already calls `dropzeros!`, so structural zeros never reach the sparse form, and the variable stays registered in the variable map.
2. No regression test for the zero/cancelling-term case above. The new tests cover dense NPP, diagonal, off-diagonal ordering, affine, constant, and sense, but not the input that exposes finding 1. Adding one would have caught the divergence and would lock in whichever behavior is chosen.

Questions:
- `QUBOTools.backend` returns `copy(model.qubo_backend_cache)` on every call. The test confirms the returned object is not identical to the cache, but is `copy` deep enough that a caller mutating the returned model's form/solution cannot corrupt the cached one across repeated `backend(model)` calls? If `copy` is shallow over the sparse forms this is fine for read-only use but worth a one-line confirmation.

Intent vs issue #183:
- The PR faithfully implements the issue's scoped first pass (remove ToQUBO-side MOI/PBO/MOI overhead, cache the backend, preserve the generic path). The issue explicitly defers the squared-affine front-end to a separate follow-up, so `Closes #183` is appropriate. Acceptance criteria are largely met except the exact "same as the current path" guarantee, which finding 1 narrows.

Tests run:
- Fetched base and head with git; merge-base confirmed at 152bae8; diff totals match (189/80, 6 files).
- `julia --project -e 'using Pkg; Pkg.test()'` on a clean head checkout: 1639/1639 pass (Julia 1.10.11), matching the PR description.
- Reproduced the variable-set divergence with a standalone MOI script (explicit zero term: cache n=2 vs parse n=3; cancelling terms: same; absent variable: both n=3).

Merge readiness:
- No blocking issues; tests pass on a clean checkout and the design is correct for canonical input. I would address finding 1 (and add its test) before merge, since it is a real regression versus the previous parse-based backend for non-canonical objectives, but it does not block on its own. The formal verdict (approval or change request) must come from another maintainer, as I am the author.

src/compiler/compiler.jl
```diff
@@ -148,127 +148,156 @@ function _copy_qubo_variables!(model::Virtual.Model{T}) where {T}
...
 end
 
-function _qubo_term(x::VI, y::VI)
-    if x.value <= y.value
-        return PBO.Term{VI}(VI[x, y])
+function _add_or_drop!(terms::Dict{K,T}, key::K, value::T) where {K,T}
+    new_value = get(terms, key, zero(T)) + value
+
+    if iszero(new_value)
+        delete!(terms, key)
```
bernalde, line 159: Nonblocking (regression vs fallback): this `delete!` removes a linear entry whenever its running sum reaches zero. `_empty_qubo_linear_terms` pre-seeds every variable at zero, so a variable whose affine/linear coefficient is explicitly zero (e.g. `ScalarAffineTerm(0.0, z)`) or cancels to zero gets dropped from the cached backend's variable set. The parse fallback `QUBOTools._extract_bool_model` (ext/QUBOTools_MOI/model_parser.jl:146) seeds `L = Dict(xi => 0 for xi in V)` and never deletes, so it keeps the variable. Reproduced: such an objective gives `QUBOTools.backend(model)` 2 variables via the cache vs 3 via the parse path, silently dropping the problem variable.

Fix: for linear terms, accumulate without deleting (match `_extract_bool_model`: `terms[key] = get(terms, key, zero(T)) + value`). `_build_sparse_forms` already calls `dropzeros!`, so structural zeros never reach the sparse form while the variable stays in the variable map. A regression test with an explicit-zero/cancelling affine term should accompany the fix.

bernalde:
Review follow-up pushed.

Commits pushed:
- `07ebdde` / `07ebdded4991ac511d749dd2c74609fcaa824239`: `fix: preserve zero-linear QUBO variables (#185)`

Main changes:
- Preserved pre-seeded linear entries in the cached QUBOTools backend when affine or diagonal-linearized terms are explicitly zero or cancel to zero.
- Added a regression test covering explicit-zero and cancelling affine terms, and checking cached backend output against target-model parsing.

Tests run:
- Focused compiler-copy test before fix: reproduced the failure, cached backend had 2 variables while target parsing had 4.
- Focused compiler-copy test after fix: passed, `78/78`.
- `git diff --check`: passed.
- `julia --project -e 'using Pkg; Pkg.test()'`: passed, `1643/1643`.

- None.

Remaining risks or follow-up:
- Post-push CI restarted after `07ebdde`; current snapshot has the label check passed and CI/docs pending.

src/compiler/compiler.jl
bernalde: Addressed in https://github.com/JuliaQUBO/ToQUBO.jl/commit/07ebdded4991ac511d749dd2c74609fcaa824239. Linear and diagonal-linearized backend terms now accumulate without deleting zero-valued entries, preserving the cached backend variable map. Added `cached backend preserves zero-linear variables` in `test/unit/compiler/copy.jl`; focused copy tests and full `Pkg.test()` pass.

bernalde:
Follow-up on the open question from my earlier review: whether `copy(model.qubo_backend_cache)` in `QUBOTools.backend` (src/wrapper.jl) is deep enough that a caller mutating the returned model cannot corrupt the cached one across repeated `backend(model)` calls.

I traced `Base.copy(::QUBOTools.Model)` (QUBOTools `src/library/model/model.jl`): it copies the variable map, copies the form (which in turn copies the underlying sparse `L` and `Q` arrays via `copy(data(...))`), and deepcopies the metadata. So mutating the coefficients, sparsity structure, variable map, or metadata of a returned backend does not touch the cache. The question is effectively answered: the copy is adequately deep for read-only consumers, which is how `backend` is used.

Recommended way to tackle it, in order of preference:

1. Within this PR, no production code change. Add one small regression test that calls `QUBOTools.backend(model)` twice, mutates the first result (e.g. zero out a coefficient or alter the offset), and asserts the second call and `model.qubo_backend_cache` are unchanged. That locks in the immutability guarantee against future `copy`/cache refactors. This is the lightest option and keeps the guarantee from silently regressing.

2. Defer to a follow-up issue if you prefer to keep this PR minimal. The issue would track adding the cache-immutability regression test and, if desired, an internal note in `backend` documenting that the cache must only ever be handed out via `copy`. There is no correctness defect to fix, so this is low priority.

I would not open a separate issue unless option 1 is explicitly out of scope; the test is small and belongs with the code it protects. Either way this is not blocking.

bernalde:
Review follow-up pushed.

Commits pushed:
- `657fbc3` / `657fbc393f13e8792080a791c6b07456bb8cd64d`: `test: cover QUBO backend cache copies (#185)`

Main changes:
- Added a wrapper regression test that calls `QUBOTools.backend(model)`, mutates the returned backend's sparse linear/quadratic data and metadata, then verifies a fresh backend and `model.qubo_backend_cache` remain unchanged.
- No production code changes were needed.

Tests run:
- Focused cache-copy regression check: passed.
- `git diff --check`: passed.
- `julia --project -e 'using Pkg; Pkg.test()'`: passed, `1650/1650`.

- None.

Remaining risks or follow-up:
- Post-push CI restarted after `657fbc3`; current snapshot has the label check passed and CI/docs pending.

PR #191, opened 2026-06-25
Reduce NPP QUBO fast-path compiler overhead

### LINKED ISSUE #190: Profile residual NPP gap after dense QUBO fast path
## Context

Issue #183 was resolved by PR #185, which shipped in ToQUBO v0.5.1. The updated benchmark confirms that the dense QUBO fast path materially improved the NPP-like case, so this should be treated as a follow-up rather than a reopening of #183.

Related work:

- #183
- #185
- #189
- Benchmark repository: `JuliaQUBO/ToQUBO-benchmark`
- Benchmark commit with recovered CI results and refreshed plots: `JuliaQUBO/ToQUBO-benchmark@5edbdc8`
- Benchmark workflow run: https://github.com/JuliaQUBO/ToQUBO-benchmark/actions/runs/28134539140

Note: the benchmark computation completed successfully in workflow run `28134539140`, but the workflow failed in `Save Results` because the `GITHUB_TOKEN` could not push generated files to protected `main`. The ToQUBO timing rows in `5edbdc8` were recovered from that CI log, so they are at printed CI-log precision.

## Current benchmark result

Stack:

```text
Julia: 1.12.6
JuMP: 1.30.1
ToQUBO: 0.5.1
QUBOTools: 0.15.1
```

For NPP at `n = 1000`:

```text
Amplify total: 0.106837511 s
ToQUBO total:  0.158 s
  JuMP/model:  0.088 s
  compiler:    0.070 s
  extraction:  0.000 s at printed CI-log precision
PyQUBO total:  0.784842730 s
```

For TSP at `n = 10000`:

```text
ToQUBO total:  1.601 s
  JuMP/model:  0.250 s
  compiler:    1.310 s
  extraction:  0.041 s
Amplify total: 5.068821192 s
```

So ToQUBO is already fastest on the dense TSP case, and the NPP gap to Amplify is now much smaller: roughly `0.051 s` at `n = 1000`.

## Why this follow-up exists

The original #183 target was to remove repeated dense quadratic materialization and dictionary-copy overhead in the already-QUBO path. That appears to have worked.

The remaining NPP gap is now small enough that we need a fresh profile on ToQUBO v0.5.1 before deciding what to optimize next. The remaining `0.158 s` appears split between:

- JuMP/model construction: `0.088 s`
- ToQUBO compile path: `0.070 s`

There are two plausible outcomes:

1. The remaining ToQUBO compiler time still has avoidable overhead, in which case the next implementation should target the v0.5.1 dense QUBO fast path.
2. Most of the remaining practical gap is now JuMP dense-expression construction, in which case beating Amplify on this syntax may require a separate front-end/API design for squared-affine or direct coefficient construction.

This issue is for measuring that boundary clearly and implementing the smallest remaining useful ToQUBO-side improvement.

## Suggested investigation

Profile the current released path on ToQUBO v0.5.1 / QUBOTools v0.15.1 for the benchmark NPP model:

```julia
@objective(model, Min, ((2x .- 1)'s)^2)
```

Use at least:

- phase timings matching the benchmark CSV fields: `model_time`, `compiler_time`, `convert_time`, `total_time`;
- a sampler profile focused on the `optimize!(model)` portion;
- allocation counts for the compile path if practical;
- a comparison against a direct dense coefficient construction baseline.

Useful benchmark reference:

```text
benchmark/ToQUBO/problems.jl
benchmark/ToQUBO/results.npp.csv
benchmark/ToQUBO/results.tsp.csv
```

## Acceptance criteria

- Add a current v0.5.1 profile summary to this issue or the fixing PR.
- Identify whether the remaining NPP gap is dominated by ToQUBO compilation or JuMP model/expression construction.
- If ToQUBO compilation remains actionable, reduce NPP `compiler_time` at `n = 1000` from about `0.070 s` toward `< 0.020 s`, without weakening correctness or bypassing the public JuMP/MOI behavior.
- If the remaining gap is primarily JuMP expression construction, document that clearly and split any squared-affine/direct-coefficient API proposal into a separate design issue.
- Preserve the correctness coverage introduced for the dense QUBO fast path, including diagonal terms, off-diagonal ordering, affine terms, constants, sense, and variable map consistency.
- Re-run `JuliaQUBO/ToQUBO-benchmark` after the change and compare against Amplify on NPP.

## Non-goals

- Do not reopen #183 unless PR #185 is found to have missed its original acceptance criteria or introduced a regression.
- Do not change benchmark semantics just to win the benchmark.
- Do not weaken correctness checks or skip dense-QUBO edge cases.

bernalde:
## Summary

Closes #190.

This PR reduces the residual NPP compiler overhead in the already-QUBO fast path by avoiding the large intermediate `Dict{Tuple{VI,VI},T}` cache build. The fast path now accumulates backend linear and quadratic data into sparse-array constructor inputs directly, while still writing the target MOI `ScalarQuadraticFunction` and preserving the existing public `QUBOTools.backend(model)` path.

It also adds coverage for duplicate, reversed, and canceling quadratic terms so the direct sparse summation path stays equivalent to parsing the target MOI model.

## Profile Summary

Local profile stack:

- Julia 1.10.11
- JuMP 1.30.1
- ToQUBO 0.5.1 checkout
- QUBOTools 0.15.1

Issue baseline at `n = 1000` from the live benchmark CSV was:

- `model_time = 0.088 s`
- `compiler_time = 0.070 s`
- `convert_time = 0.000 s` at printed precision
- `total_time = 0.158 s`

Sampler profile before this change showed `optimize!` dominated by `_copy_qubo_objective!`, especially the quadratic-term dictionary accumulation and the subsequent `QUBOTools.Model` dictionary-to-sparse conversion.

After this change, a local phase profile on the NPP-like model produced these stable post-warmup timings:

- JuMP syntax profile: `compiler_time ~= 0.029 s`, `convert_time ~= 0.0036 s`
- Direct MOI coefficient baseline: `compiler_time ~= 0.026 s`, `convert_time ~= 0.0052 s`

Running the benchmark repository's exact NPP problem function with this local checkout gave post-warmup `n = 1000` compiler times around `0.047-0.051 s`. On this Julia 1.10 machine, the same run had `model_time ~= 0.18-0.20 s`, so the remaining practical gap on the public JuMP syntax is now dominated by JuMP model/expression construction rather than backend extraction. The live Amplify CSV row at `n = 1000` remains `0.106837511 s`.

The squared-affine/direct-coefficient front-end idea should remain separate design work; this PR only optimizes the existing public JuMP/MOI behavior.

## Tests

- `git diff --check`
- `julia --project=. -e 'push!(LOAD_PATH, joinpath(pwd(), "test")); using Test; using JuMP; const MOIU = MOI.Utilities; const VI = MOI.VariableIndex; const CI{F,S} = MOI.ConstraintIndex{F,S}; using QUBODrivers; using LinearAlgebra; using TOML; using ToQUBO; using ToQUBO: Attributes, Encoding; import QUBOTools; import PseudoBooleanOptimization as PBO; function MOIU.map_indices(::Function, arch::QUBOTools.AbstractArchitecture); return arch; end; function MOIU.map_indices(::Function, quad::PBO.QuadratizationMethod); return quad; end; include("test/unit/compiler/copy.jl"); test_compiler_copy()'`
  - `85/85` passing
- `julia --project=. -e 'push!(LOAD_PATH, joinpath(pwd(), "test")); include("test/runtests.jl")'`
  - `1665/1665` passing
- Local benchmark NPP function from `JuliaQUBO/ToQUBO-benchmark` using this checkout as the active Julia project.

## Branch Hygiene

- Base branch: `main`
- Source branch: `fix/issue-190-npp-compiler-time`
- Source branch point: `ae94c0f75d08010c2f9c155bfa28b2b506cf0083` (`origin/main`)
- Stacked status: not stacked
- Prerequisite PRs: none


Project.toml
```diff
@@ -8,6 +8,7 @@ MathOptInterface          = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 PseudoBooleanOptimization = "c8fa9a04-bc42-452d-8558-dc51757be744"
 QUBOTools                 = "60eb5b62-0a39-4ddc-84c5-97d2adff9319"
 Random                    = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SparseArrays              = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
```
bernalde, line 11: Blocking (compat) plus a nonblocking registry item, both anchored here:

- Nonblocking: this new `SparseArrays` dep has no `[compat]` entry. General registry AutoMerge requires a bound for every dependency, so the next release will be rejected. Add `SparseArrays = "1"`.
- Blocking: the real location is line 24, `QUBOTools = "0.11, 0.12, 0.13, 0.14, 0.15"`. This PR newly uses `QUBOTools.Form`, `SparseLinearForm`, `SparseQuadraticForm`, `VariableMap{VI}(::Vector)`, and `Model(::VariableMap, ::Form)`, none used by ToQUBO before. Verified only against 0.15.1. Either confirm those symbols exist across 0.11-0.14 or tighten the lower bound.

src/compiler/compiler.jl
```diff
@@ -148,55 +149,49 @@ function _copy_qubo_variables!(model::Virtual.Model{T}) where {T}
...
+    target_variables = _target_variables(variable_map)
+    n = length(target_variables)
+    L = SparseArrays.sparsevec(linear_indices, linear_values, n)
+    Q = SparseArrays.sparse(quadratic_rows, quadratic_cols, quadratic_values, n, n)
+
+    SparseArrays.dropzeros!(L)
+    SparseArrays.dropzeros!(Q)
+
+    backend_variable_map = QUBOTools.VariableMap{VI}(target_variables)
+    backend_form = QUBOTools.Form{T}(
```
bernalde, line 184: Nonblocking (maintainability): this block re-implements QUBOTools' private `_build_sparse_forms` almost verbatim. The `i < j` upper-triangular split (lines 290-301), the `sparsevec`/`sparse` construction, and the `dropzeros!` calls all mirror `QUBOTools/src/library/model/model.jl`; this code now also depends on the internal `Form`/`SparseLinearForm`/`SparseQuadraticForm` normal-form types. If QUBOTools changes its normal form (triangle convention, scale semantics, drop behavior), this hand-rolled copy diverges silently.

What to do about it: keep the fast path, but make the dependency explicit and guarded rather than implicit. Concretely: (1) add a comment here naming `QUBOTools._build_sparse_forms` as the source of truth and the `_dense_qubo_data(backend) == _dense_qubo_data(parsed_backend)` test as the equivalence guard; (2) open a QUBOTools issue to expose a public constructor that takes COO triples or prebuilt sparse `L`/`Q` directly, so ToQUBO can depend on a stable contract and delete this replica; (3) until then, the equivalence test is what protects us, so it should not be weakened. This is the maintainability tradeoff for the speedup, and it is acceptable as long as it is documented and the compat bound is honest.

src/compiler/compiler.jl
```diff
@@ -148,55 +149,49 @@ function _copy_qubo_variables!(model::Virtual.Model{T}) where {T}
...
+    quadratic_values::Vector{T},
     offset::T,
 ) where {T}
-    model.qubo_backend_cache = QUBOTools.Model{VI,T,Int}(
-        linear_terms,
-        quadratic_terms;
-        offset,
+    target_variables = _target_variables(variable_map)
+    n = length(target_variables)
+    L = SparseArrays.sparsevec(linear_indices, linear_values, n)
```
bernalde, line 175: Nonblocking: `L`/`Q` are indexed by `y.value`, but `QUBOTools.VariableMap{VI}(target_variables)` (line 178) sorts the variables and assigns each its sorted rank, so `index(vm, y)` equals `y.value` only when target variables have contiguous values `1..n`. That holds today because the fast path creates them sequentially, but it is undocumented and unguarded; a future non-contiguous target map would misalign the Form or throw `BoundsError` from `sparsevec(..., n)`, and the new equivalence test (small contiguous cases) would not catch it. Add a comment stating the invariant, or build the indices through the VariableMap rank instead of `y.value`.

bernalde (commented):
Review by a maintainer who is also the PR author, so GitHub permits only a COMMENT event here; the formal approval or change request must come from another maintainer. Full findings below. There is one blocking issue, so: I would not merge this until the blocking issue below is addressed.

Setup: fetched `origin/main` and `pull/191/head` (`feef214`); merge-base is `ae94c0f` (matches the PR's stated source branch point). Diff against the merge-base is 129 additions / 65 deletions across 3 files, matching the PR's reported totals. Reviewed on a clean checkout of the head.

The change is sound in substance: it removes the `Dict{Tuple{VI,VI},T}` intermediate and accumulates COO inputs directly, then assembles the QUBOTools backend Form. The math matches MOI conventions (diagonal `c/2` into linear, off-diagonal `c` into the strict upper triangle) and the duplicate/cancel handling is correct because `sparsevec`/`sparse` sum duplicates and `dropzeros!` removes cancellations. The new equivalence test against `QUBOTools.Model{Float64}(model.target_model)` is the right guard. Replacing `MOI.set(..., ObjectiveFunction, f)` with a direct field write is safe: the previous `MOI.set` copied the term vectors, but `f` is built from fresh locals that are not aliased elsewhere, so dropping the copy is the intended optimization, not a bug.

Blocking

1. QUBOTools compat range no longer matches the API the PR uses. `Project.toml` keeps `QUBOTools = "0.11, 0.12, 0.13, 0.14, 0.15"`, but this PR is the first place ToQUBO reaches into QUBOTools' lower-level normal-form types: `QUBOTools.Form`, `QUBOTools.SparseLinearForm`, `QUBOTools.SparseQuadraticForm`, `QUBOTools.VariableMap{VI}(::Vector)`, and the two-arg `Model(::VariableMap, ::Form)` constructor. None of these were referenced anywhere in ToQUBO before (confirmed by `git grep` against `origin/main`). The old code used the higher-level dict constructor `QUBOTools.Model{VI,T,Int}(linear_terms, quadratic_terms; ...)`. I verified the change only against the installed QUBOTools 0.15.1; I cannot confirm these symbols exist with the same signatures in 0.11-0.14. If they do not, a resolve against any allowed older version will fail to precompile (UndefVarError) or error at runtime for users not pinned to 0.15. Fix: either confirm these exact symbols exist across the whole declared range and say so, or tighten the lower bound to the first QUBOTools version that exposes them (the only version actually tested here is 0.15).

Nonblocking

1. The new `_cache_qubo_backend!` re-implements QUBOTools' private `_build_sparse_forms` almost line for line (same `i < j` upper-triangular split, same `sparsevec`/`sparse` + `dropzeros!`, same Form assembly). Detail and a recommendation are in the inline comment on `compiler.jl`. The equivalence test mitigates silent drift, but the duplication should be acknowledged in code and ideally retired by a public QUBOTools COO/sparse constructor.

2. `SparseArrays` is added to `[deps]` with no `[compat]` entry. It is a stdlib, but the General registry AutoMerge requires a compat bound for every dependency, so this will block registration of the next release. Add `SparseArrays = "1"` (and, for consistency with the rest of the file, `Random = ...`-style stdlib bounds are already omitted, so at minimum add the one this PR introduces).

3. The backend Form is indexed by `y.value` while `QUBOTools.VariableMap{VI}(target_variables)` independently sorts the variables and assigns each its sorted rank. These agree only because the QUBO fast path creates target variables with contiguous values `1..n` (via `QUBOModel.add_constrained_variable`). The invariant is real today but undocumented and unguarded: if target variables ever became non-contiguous, `sparsevec(indices, vals, n)` would misalign or throw `BoundsError`, and the new equivalence test (small, contiguous cases only) would not catch it. Add a short comment stating the invariant, or index through the VariableMap rank rather than `y.value`.

Questions

1. The PR says `Closes #190`. Issue #190's actionable target was to reduce NPP `compiler_time` at `n = 1000` from ~0.070 s toward `< 0.020 s`, with an explicit alternative: if the residual is dominated by JuMP expression construction, document that and split any front-end work into a separate issue. The PR reports ~0.029 s compiler time (improved, but not under 0.020 s) and concludes the remainder is JuMP-side, which is one of the sanctioned outcomes, so closing is defensible. But two acceptance criteria look only partially met: the `< 0.020 s` target was not reached, and "re-run `JuliaQUBO/ToQUBO-benchmark` and compare against Amplify on NPP" appears to have been a local function run rather than a CI benchmark re-run. Is `Closes #190` intended, or should this be `Refs #190` with the benchmark re-run and the squared-affine follow-up issue tracked separately?

Tests run (clean head checkout, Julia 1.10.11, QUBOTools 0.15.1):
- Targeted `test/unit/compiler/copy.jl::test_compiler_copy()`: 85/85 pass.
- Full `test/runtests.jl`: pass, exit 0, no failures/errors/broken. The two `Indicator Constraint` entries marked `None` are pre-existing optional-solver skips, not introduced here.

Merge readiness: not ready as-is. The implementation and tests are correct against QUBOTools 0.15.1, but the declared compat range advertises support for versions whose APIs this PR has not been verified against. Resolve the blocking compat item; the nonblocking items are quality/maintainability and do not by themselves block. I would not merge this until the blocking issue above is addressed.

bernalde:
Pushed review-feedback fixes in `1eabb32`.

Main changes:

- Tightened QUBOTools compat from `0.11, 0.12, 0.13, 0.14, 0.15` to `0.15` in the root, test, and docs projects because this PR now depends on QUBOTools 0.15 normal-form APIs.
- Added `SparseArrays = "1"` to root `[compat]`.
- Added code comments documenting the contiguous target-variable invariant and the temporary mirror of `QUBOTools._build_sparse_forms`.
- Opened JuliaQUBO/QUBOTools.jl#115 for a public COO/sparse constructor so ToQUBO can later remove the local sparse-form mirror.

Tests run:

- `git diff --check`: pass
- Targeted compatibility test: `22/22` pass
- Targeted compiler copy test: `85/85` pass
- Full local suite: `1657/1657` pass

- None declined. I kept `Closes #190`; this PR still implements the ToQUBO-side improvement and documents the remaining JuMP-side boundary, while the QUBOTools constructor cleanup is tracked separately in JuliaQUBO/QUBOTools.jl#115.

Remaining risks or follow-up:

- CI has not completed yet for `1eabb32`.
- The local sparse-form mirror should be replaced once QUBOTools exposes a stable public constructor.

Project.toml
bernalde: Addressed in 1eabb32b192d63bfb637a5232c728a956876eafb: tightened QUBOTools compat to 0.15 and added SparseArrays compat. Updated the root, test, and docs project compat plus test/unit/compat.jl.

src/compiler/compiler.jl
bernalde: Addressed in 1eabb32b192d63bfb637a5232c728a956876eafb: added the QUBOTools._build_sparse_forms mirror comment and opened JuliaQUBO/QUBOTools.jl#115 for a public COO/sparse constructor. The copy tests continue to guard cached-backend parity.

src/compiler/compiler.jl
bernalde: Addressed in 1eabb32b192d63bfb637a5232c728a956876eafb: added a comment documenting that fast-path target variables come from an empty QUBOModel with values 1:n, matching QUBOTools.VariableMap sorted ranks.

PR #210, opened 2026-07-16
docs: explain constraint reformulation

bernalde:
## Summary

Closes #209.

- document the compiler's scalar equality, inequality, interval, sign-definite, and quadratization behavior in the encoding booklet
- explain how finite continuous-slack resolution introduces a residual penalty floor, and cross-link the corresponding compiler settings
- remove the misleading empty constraint-encoding stub and its dead commented include
- add a focused regression guard for the documentation and source-layout contract

## Related Work

- #179 provides the existing representation-error derivation linked by this change
- #141 and #178 provide the existing penalty-setting and slack-provenance documentation used by the new cross-links
- #205 remains the sampling-feasibility symptom motivating the explanation
- #208 continues to own analysis of slack resolution versus penalty bias
- #207 continues to own any future slack-free encoding implementation

This PR documents current behavior and does not claim to resolve #205, #207, or #208.

## Verification

- regression guard before the implementation: `0 passed, 10 failed`
- `julia --project=test -e 'using Test; include("test/unit/constraint_reformulation_docs.jl"); test_constraint_reformulation_docs()'` — `10/10` passed
- `julia --project -e 'using Pkg; Pkg.test()'` — `1693/1693` passed on Julia 1.10.11
- `julia --project=docs docs/make.jl --skip-deploy` — doctests, cross-references, document checks, and rendering passed
- `git diff --cached --check`

## Branch Hygiene

- Base branch: `main`
- Source branch point: `664a77fd2c6fb51430a139753c79608d27fc1407` (`origin/main`)
- Stacked status: standalone
- Prerequisite PRs: none
- Open-PR file overlap: none at push time

Issue #208, opened 2026-07-11
Analyze discretized-slack granularity vs. feasibility/penalty bias


bernalde:
## Context

Follow-up analysis from a discussion with @K-A-McDonald about continuous-constraint
encoding. This is a likely **mechanistic explanation** for the penalty-tuning
symptom reported in #205, and complements (does not duplicate) it.

## The problem

When ToQUBO binary-encodes a **continuous** slack at resolution `δ`, the residual
`a'x - b + z` generally cannot be driven exactly to zero — the slack grid does not
land on the value needed to cancel the residual. The penalty `(residual)²` is then
strictly positive **even at genuinely feasible continuous points**.

Consequences:

- Feasible points carry a residual penalty floor `~O(δ²)`, so the penalty weight `ρ`
  must be inflated enough that this floor still dominates any objective gain from
  cheating — which is exactly the manual `1e7` inflation @K-A-McDonald needed in #205.
- The induced bias interacts with the uniform penalty scaling
  (ToQUBO set every encoding penalty to `26101` in that model), so constraints with
  tighter effective granularity are under-penalized relative to others.

The paper's slack-free penalty (see companion feature issue) sidesteps this because
it never requires exact cancellation.

## Why this matters

If confirmed, the equality/inequality penalty for continuous constraints should be
chosen relative to the slack encoding resolution, not uniformly — and the residual
should arguably be relaxed to a tolerance band matched to `δ` rather than requiring
exact zero. This would reduce the need for manual penalty inflation.

## Proposed work

- Verify against the code how the constraint encoding penalty is currently chosen
  relative to slack/variable encoding resolution (`src/compiler/penalties.jl`,
  `src/compiler/constraints.jl`).
- Quantify the feasible-point penalty floor as a function of `δ` on the #205 model.
- Evaluate options:
  - scale the constraint penalty per-constraint using its slack resolution, and/or
  - relax the residual to a `δ`-matched tolerance band before squaring.
- Confirm whether this explains #205, and whether it closes the manual-inflation gap.

## Acceptance criteria

- A documented relationship between slack encoding resolution and the required
  penalty weight.
- Either an automatic per-constraint penalty scaling / tolerance relaxation, or a
  documented recommendation, that removes the need for manual inflation on #205.

## Notes / references

- Related: #205 (empirical symptom), #137 (penalty heuristic parameterization),
  #172 (objective-range exact penalty policy).
[...]

Issue #207, opened 2026-07-11
Add a slack-free inequality penalty encoding method


bernalde:
## Context

This issue comes out of a discussion with @K-A-McDonald about how ToQUBO encodes
**real-valued (continuous) inequality constraints**, and how that compares to the
slack-free penalty formulation used in the literature (e.g. eq. 5 of
[Power-network optimization, IEEE ACCESS 2023](https://doi.org/10.1109/ACCESS.2023.3312997)).

### How ToQUBO handles inequalities today

For an inequality `a'x - b ≤ 0`, the compiler
([`src/compiler/constraints.jl`](../blob/main/src/compiler/constraints.jl)):

1. adds a bounded slack `z ≥ 0` to form the equality `a'x - b + z = 0`, then
2. squares the residual into a quadratic penalty `(a'x - b + z)²`.

Integer and continuous constraints go through the *same* path. Because QUBO is
binary, a continuous slack cannot be represented exactly, so `variable_ℝ!`
**binary-encodes the slack** over a bounded interval at a user-chosen resolution
(`SlackVariableEncodingBits` / `..ATol`). ToQUBO therefore does not need the
paper's one-sided approximate penalty — but it pays for it in extra qubits and a
discretization error.

## Why this matters

The whole motivation for the paper's slack-free penalty is **qubit economy**: each
binary-encoded continuous slack costs `~log2(range/resolution)` extra qubits *per
inequality*, plus the couplings to embed them on hardware. For NISQ/annealing
targets this is often the dominant cost.

ToQUBO already has a pluggable `Attributes.constraint_encoding_method` with
`QuadraticPenalty` as one implementation, so a slack-free alternative fits the
existing architecture as an additional method rather than a rewrite.

## Proposed work

- Add a **slack-free inequality penalty** as an alternative
  `constraint_encoding_method`, e.g. a `max(0, g)²`-style penalty and/or the
  paper's one-sided approximation (eq. 5).
- Let users select per-constraint (same mechanism as the existing method attribute).
- Document the tradeoff: slack-free saves qubits but introduces a modeling
  approximation / non-quadratic exact form (needs quadratization), whereas the
  current slack method is exact-up-to-slack-resolution but costs qubits.
- Tests comparing feasibility/qubit-count of both methods on a small MI model.

## Acceptance criteria

- A slack-free penalty method is selectable via the constraint encoding attribute.
- Docs explain when to prefer it over the slack reformulation.
[...]

PR #206, opened 2026-07-07
docs: clarify Neal penalty tuning

bernalde:
## Summary

Refs #205.

- document finite-run stochastic sampler penalty tuning for `DWave.Neal`
- recommend targeted `ConstraintPenaltyScale()` overrides for frequently violated source constraints instead of global `PenaltyScale()` or fixed hints everywhere
- add a docs unit assertion so the Neal tuning guidance stays present

## Reproduction Notes

The linked reproducer repository from #205, `ZOMEGA-Group/QUBO_UC`, was not accessible from this environment through `gh repo clone` or GitHub search. I therefore used an isolated `/tmp/toqubo-205-repro` Julia project with local ToQUBO plus:

- `DWave v0.7.6`
- `JuMP v1.30.1`
- `QUBODrivers v0.6.5`
- `QUBOTools v0.16.0`

A small UC-style mixed-integer fixture was scaled so ToQUBO inferred the same uniform equality penalty reported in the issue: `26101.0`.

Read-weighted `DWave.Neal` results over 10 seeds and 100 reads per seed showed:

- default penalties: mean feasible rate about `0.895` with Neal default sweeps; demand equality violations in `5%` to `14%` of reads
- fixed `1e7` equality hints: mean feasible rate about `0.997`; demand equality violations `0%`
- targeted `ConstraintPenaltyScale()` on the demand equalities with scale `400`: demand penalties about `1.04404e7`, mean feasible rate about `0.997`, and demand equality violations `0%`
- global `PenaltyScale()` with the same scale was worse in this fixture because it also scaled unrelated penalties and changed the sampler landscape

This supports treating #205 as finite-sampler penalty tuning guidance rather than changing the default exact-penalty policy.

## Local Checks

- `julia --project=test -e 'push!(LOAD_PATH, pwd()); using Test; include("test/unit/docs.jl"); test_docs()'`
- `git diff --check`

## Branch Hygiene

- Base branch: `main`
- Source branch point: `origin/main`
- Stacked status: standalone
- Prerequisite PRs: none

PR #211, opened 2026-07-17
analysis: quantify slack-resolution penalty contrast

bernalde:
## Summary

Refs #208.

- add an executable public fixture that measures continuous-slack grid spacing, feasible residual floors, penalty contrast, and inferred QUBO ranking across one to six bits;
- pin a compiler audit and seeded Neal aggregate for the canonical #205 model without copying its access-controlled source;
- add a generated Markdown report and a golden regression guard so the public evidence remains reproducible.

This is the first, analysis-only slice agreed in [the issue discussion](https://github.com/JuliaQUBO/ToQUBO.jl/issues/208#issuecomment-5003349226). It deliberately does not change compiler behavior.

## Findings and decision

- A continuous binary slack on `[0, S]` has grid spacing `Delta = S / (2^n - 1)` and feasible squared-residual floor at most `Delta^2 / 4`, but safe penalty sizing depends on the contrast `p_I - p_F`, not on resolution alone.
- The one-bit public fixture has `p_I < p_F`; no positive penalty multiplier can repair that ordering. This falsifies a universal resolution-only scaling rule.
- The canonical #205 model has 36 binary source variables. Its c2-c4 families are equalities with no generated slack, and its inequality slacks use integral expansions under the default `Discretize() == true` path. The proposed continuous-slack mechanism is therefore absent from that model.
- Across 100,000 seeded Neal reads per configuration, c2-c4 scale 400 produces a 0.156% feasible-read rate (close to 0.141% for the fixed `1e7` hint), while scaling only slack-bearing inequalities produces no feasible read.

The evidence supports a follow-up documentation PR explaining the `Discretize()` boundary and penalty-contrast condition. It does not justify a resolution-aware compiler implementation for #205; such a change needs separate design evidence from genuinely continuous-slack models.

## Provenance

- ToQUBO base: `ed438ebe8ad6d39873c52894ddcc25a2825cb62c`
- canonical model: access-controlled `ZOMEGA-Group/QUBO_UC`, branch `QUBO_feasibility_tests`, commit `707b30a527d33da3df0cdeb3ca336d55a28ca90a`
- canonical source copied here: no
- public benchmark: `benchmarks/slack_resolution_analysis.jl`
- generated report: `benchmarks/reports/slack_resolution_analysis.md`

The canonical sweep used local Neal 0.6.0 and dimod 0.12.21, seeds 1-10, 10,000 reads per seed, and 100 sweeps.

## Related work

- #205 remains the empirical sampler-symptom issue.
- #206 supplied targeted `ConstraintPenaltyScale()` guidance.
- #207 separately owns slack-free inequality encoding.
- #210 documented the qualitative continuous-slack mechanism.
- #173 established the objective-range penalty policy.

## Verification

- `JULIA_LOAD_PATH="@:test:@stdlib" julia +1.10 --project=. benchmarks/slack_resolution_analysis.jl`
- focused guard: 29/29 passing
- `julia +1.10 --project=. -e 'using Pkg; Pkg.test()'`: 1,723/1,723 passing using the local installed environment
- `git diff --cached --check`

## Branch hygiene

- base: `main` at `ed438ebe8ad6d39873c52894ddcc25a2825cb62c`
- head: `analysis/issue-208-slack-resolution`
- standalone analysis PR with no prerequisite PR


bernalde (commented):
Maintainer review — analysis-only PR. I authored this PR, so GitHub only allows a `COMMENT` verdict from this account; branch protection on `main` requires **0** approving reviews, so no separate approval is needed to merge.

This is an evidence/analysis slice for #208 (correctly linked `Refs #208`, not `Closes` — #208 is `needs-design` and stays open). It adds no compiler behavior. I verified the reproducible portion end-to-end rather than trusting the committed numbers.

What I verified:

- **Public fixture is genuinely computed, not pinned.** `analyze_bits` builds a real `ToQUBO.Optimizer` model, calls `optimize!`, and reads `reformulation_metadata` (a public API in `src/reformulation.jl`); `StableCompilation`, `SlackVariableEncodingBits`, and `Discretize` all exist. I hand-checked the fixture (`0.6(x1+x2) ≤ 1`, `min -(x1+x2)`, slack on `[0,1]`): feasible residual `-0.4`, infeasible `0.2`, `B = 1`. That yields `p_F = 0.16`, `p_I = 0.04`, contrast `-0.12`, required `ρ` impossible at 1 bit; and `p_F = 0`, required `ρ = 25` at 4 bits (z*=0.4 lands on the 16-point grid). Both match the committed report table.
- **Golden guard is a real reproducibility check.** The test regenerates the full Markdown via `write_markdown_report` and compares byte-for-byte against `benchmarks/reports/slack_resolution_analysis.md`. CI passing on Ubuntu+Windows × Julia 1+1.10 confirms the live-computed table and the number formatting are deterministic cross-platform.
- **The math checks out.** `binary_grid_spacing = span/(2^n-1)`, `maximum_squared_residual_floor = Δ²/4`, and `required_penalty = B/(p_I-p_F)` (Inf when contrast ≤ 0) are all consistent with the report's derivation and with the `p_F ≤ Δ²/4` bound asserted per row.
- **Issue linkage is accurate.** The conclusion (mechanism real in the `Discretize(false)` fixture, but absent from the canonical #205 model whose c2-c4 families are slack-free equalities, and controlled by penalty contrast rather than resolution alone) directly answers #208's ask, and the PR explicitly declines to justify a compiler change from this evidence.

## Blocking

None.

## Nonblocking

- The canonical compiler-audit and Neal-sweep tables (`CANONICAL_COMPILATION_AUDIT`, `CANONICAL_NEAL_RESULTS` in `benchmarks/slack_resolution_analysis.jl`) are hardcoded constants from the access-controlled `ZOMEGA-Group/QUBO_UC` model, so CI cannot re-derive them — only the `public_fixture` table is regenerated and self-checked. This is a reasonable, clearly-disclosed choice for an analysis PR; just flagging that those two tables are pinned evidence resting on the provenance record, not reproducible output. Optional: a one-line note in the report marking them as pinned inputs would help future readers. 
- In `analyze_bits` the feasible/infeasible residuals (`-0.4`, `0.2`) and objective values (`-1.0`, `-2.0`) at benchmarks/slack_resolution_analysis.jl:254-264 are hardcoded to `build_fixture`'s specific `0.6·sum(x) ≤ 1` / `min -sum(x)` definition. If `build_fixture` is later edited (e.g. a different coefficient) the slack span stays `[0,1]`, so the `@assert length(grid) == 2^bits` guard would not catch the desync — only the byte-for-byte golden guard would, and only if the report isn't regenerated in the same edit (a regenerated report would be internally consistent but wrong). Consider deriving these from the model/metadata, or add a comment tying them to the fixture. 

## Questions

None.

## Tests run

- `JULIA_LOAD_PATH="@:test:@stdlib" julia --project=. -e 'using Test; include("test/unit/slack_resolution_analysis.jl"); test_slack_resolution_analysis()'` → **29/29 passed** at head `0a2b866` (~10.5s; the runtime confirms real model compilation, not constant lookups).
- Confirmed wiring: `runtests.jl` → `test_unit()` → `test_slack_resolution_analysis()`.
- Working tree clean after the run — no generated artifacts left behind.

## Merge-readiness

- Changed-file set matches `gh pr diff --name-only` (4 files, +794/-0).
- All required status checks green for head `0a2b866` (Julia 1 / 1.10 × ubuntu / windows), plus `build` and `documenter/deploy`.
- Branch protection: `required_approving_review_count: 0`, `required_conversation_resolution: true`. This body-only review creates no threads, so it does not change merge state (`CLEAN`).

No blocking findings; the PR delivers a self-checked, reproducible analysis slice for #208 with clear provenance for the non-reproducible canonical portion. This account cannot cast a formal approval, but branch protection does not require one.

bernalde:
Addressed both maintainer-review findings in `1eae31c`.

- `fixture-coupled-constants`: `analyze_bits` now evaluates the feasible and infeasible residuals and objective values directly from the JuMP fixture. The penalty threshold and comparison energies use those derived values instead of duplicated literals. Five independent literal assertions preserve the intended fixture states.
- `pinned-canonical-tables`: the generated report now explicitly identifies the canonical compiler-audit and Neal-sweep tables as pinned evidence inputs from the access-controlled model that the public script reports but cannot re-derive.

Verification:

- focused analysis guard: 35/35 passed;
- mutation check against pre-fix head `0a2b866`: the five new provenance assertions errored because the old result exposed none of the derived fields, confirming the guard is non-vacuous;
- full suite: 1,729/1,729 passed;
- JuliaFormatter applied using the repository configuration;
- staged diff and secret/path scan clean.

Intentionally unaddressed comments: none. There are no inline review threads to reply to or resolve. No formal approval is required by branch protection; the existing review decision remains a body-only `COMMENT`.

Issue #205, opened 2026-07-07
[bug] Penalty terms need to be inflated for feasible results


K-A-McDonald:
### Summary

When passing a small mixed-integer unit commitment model through ToQUBO and solving with DWave.Neal, feasible solutions are very scarce. Only after manually inflating some of the constraint encoding penalty terms does the solver achieve usable results.

[Control.pdf](https://github.com/user-attachments/files/29743375/Control.pdf) shows the probability histogram over 100 runs with the penalties unchanged. [Manual_encoding_penalty.pdf](https://github.com/user-attachments/files/29743443/Manual_encoding_penalty.pdf) shows 100 runs with some penalties manually inflated.

https://github.com/ZOMEGA-Group/QUBO_UC/tree/QUBO_feasibility_tests

Linked is the model used to obtain the probability histograms. Penalty terms can be adjusted in Binary_det.jl; the best results were obtained by setting the encoding penalty for the equality constraints c2-4 to 1e7. These constraints are typically violated the most often. 

For this model, ToQUBO uniformly sets each encoding penalty to a value of 26101.

### Reproduction

```shell
set_attribute.(c2, Ref(ToQUBO.Attributes.ConstraintEncodingPenaltyHint()), 1e7)
set_attribute.(c3, Ref(ToQUBO.Attributes.ConstraintEncodingPenaltyHint()), 1e7)
set_attribute.(c4, Ref(ToQUBO.Attributes.ConstraintEncodingPenaltyHint()), 1e7)
```

### Expected behavior

ToQUBO should produce feasible results

bernalde:
I was able to reproduce the qualitative Neal behavior with the reported solver stack, but the linked `ZOMEGA-Group/QUBO_UC` repository is not accessible from this environment via `gh repo clone` or GitHub search, so I used a small UC-style fixture instead.

Temporary reproduction environment:

- local ToQUBO checkout
- `DWave v0.7.6`
- `JuMP v1.30.1`
- `QUBODrivers v0.6.5`
- `QUBOTools v0.16.0`

The fixture was scaled so ToQUBO inferred the same uniform equality penalty reported here: `26101.0`. Read-weighted `DWave.Neal` results over 10 seeds and 100 reads per seed showed:

- default penalties: mean feasible rate about `0.895` with Neal default sweeps; demand-equality violations in `5%` to `14%` of reads
- fixed `1e7` equality hints: mean feasible rate about `0.997`; demand-equality violations `0%`
[...]

PR #212, opened 2026-07-20
docs: explain slack-resolution penalty contrast

bernalde:
## Summary

- qualify the objective-range exact-penalty guidance when a feasible source
  assignment cannot achieve zero generated penalty;
- document the `Discretize(true)` integer-slack path and the
  `Discretize(false)` continuous-slack path;
- derive the continuous-slack floor and the required contrast condition
  `rho * (p_I - p_F) > B`, including the case where no positive penalty can
  repair the ordering;
- add a diagnostic sequence that separates slack ownership, exact-model energy
  ordering, and finite-run sampler behavior;
- update the booklet cross-reference and add a dedicated docs regression guard.

Refs #208.

## Scope and evidence

This is Part 2 of the sequential plan recorded on #208. Part 1 landed in #211
and found that continuous-slack resolution is a real mechanism under
`Discretize(false)`, but it does not explain #205: the reported c2-c4 families
are slack-free binary equalities, and that model's generated inequality slacks
are integral under `Discretize(true)`.

Accordingly, this PR documents targeted `ConstraintPenaltyScale()` tuning for
#205-style finite-run sampler pressure without presenting resolution-aware
scaling as its remedy. It changes no compiler behavior, copies no
access-controlled source or data, and leaves any automatic scaling or
tolerance-band design for a separate decision supported by genuinely
continuous-slack evidence.

Related boundaries remain unchanged:

- #206 owns targeted Neal penalty tuning for #205;
- #207 owns a possible slack-free inequality encoding;
- #210 explains the current constraint reformulations;
- #173 owns the current objective-range penalty policy.

## Verification

- Initial docs guard on unchanged documentation: 11/11 assertions failed.
- Final targeted docs guard: 13/13 assertions passed.
- `julia +1.10 --project=docs docs/make.jl --skip-deploy`: passed with the
  repository's existing local docs environment.
- `julia +1.10 --project=. -e 'using Pkg; Pkg.test()'`: 1,742/1,742 tests
  passed (1,417 unit, 325 integration) with Julia 1.10.11 and the existing
  local package cache.
- `git diff --cached --check`: passed before commit.

## Branch Hygiene

- Base: `main` at `117751958c3ce065543127f7461e8dc9ffcd107a`.
- Branch point: exact `origin/main` tip after merged Part 1 PR #211.
- Source: `docs/issue-208-slack-penalty-guidance`.
- Stacked: no; #211 is already in the base.
- Open-PR file overlap at commit time: none.

### DIFF
--- a/docs/src/booklet/4-encoding.md
+++ b/docs/src/booklet/4-encoding.md
@@ -172,16 +172,20 @@ tolerance used for bit-count inference. The selected encoding method comes from
 The required real-valued slack can fall between two values on the encoded grid.
 In that case ``g(x) \pm z`` cannot reach zero even though the original
 continuous inequality is feasible, leaving a positive squared-residual floor.
-Finer resolution reduces this representation error but consumes more target
+The floor can be zero when the required slack lands exactly on the grid, and
+grid alignment means it need not decrease strictly at every added bit. Finer
+resolution generally reduces the representation error but consumes more target
 binary variables and can change the effective penalty landscape. The bit-count
-relationship is derived in [Representation Error](@ref), and the user-facing
-settings are summarized under [Continuous Variable Resolution](@ref).
+relationship is derived in [Representation Error](@ref). The user-facing
+settings and the required feasible/infeasible penalty contrast are summarized
+under [Continuous Variable Resolution](@ref) and
+[Slack Resolution and Penalty Contrast](@ref).
 
 Sampling-feasibility behavior that motivated this documentation is tracked in
-[#205](https://github.com/JuliaQUBO/ToQUBO.jl/issues/205). The resolution and
-penalty interaction is being analyzed in
-[#208](https://github.com/JuliaQUBO/ToQUBO.jl/issues/208), while a possible
-slack-free alternative is tracked separately in
+[#205](https://github.com/JuliaQUBO/ToQUBO.jl/issues/205). The
+[issue #208 analysis](https://github.com/JuliaQUBO/ToQUBO.jl/blob/main/benchmarks/reports/slack_resolution_analysis.md)
+found that continuous-slack resolution does not explain that model's reported
+violations, while a possible slack-free alternative is tracked separately in
 [#207](https://github.com/JuliaQUBO/ToQUBO.jl/issues/207).
 
 Equality constraints use a squared residual penalty by default. The compiler
--- a/docs/src/manual/4-settings.md
+++ b/docs/src/manual/4-settings.md
@@ -54,12 +54,15 @@ where `s` is [`ToQUBO.Attributes.PenaltyScale`](@ref), `β` is
 [`ToQUBO.Attributes.PenaltyOffset`](@ref), `σ` is `1` for minimization models
 and `-1` for maximization models, `U - L` is the compiled pseudo-boolean
 objective range, and `ϵ` is the smallest positive value of the generated
-nonnegative penalty function. With the default scale `s = 1` and `β > 0`, any
-infeasible assignment is penalized by more than the largest possible objective
-improvement available within the compiled objective range. With custom scales,
-the same sufficient exactness condition is `s * (U - L + β) > U - L`; smaller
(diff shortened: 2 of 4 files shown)


bernalde (commented):
Maintainer review — documentation PR (Part 2 of the #208 plan, correctly linked `Refs #208`). I authored this PR, so GitHub only allows a `COMMENT` verdict from this account; branch protection on `main` requires **0** approving reviews, so no separate approval is needed to merge.

I verified the documentation against the source of truth rather than just reading the prose:

- **Exact-penalty qualification is correct and material.** The prior text asserted the `s = 1, β > 0` exactness guarantee unconditionally; the new text gates it on "every feasible source assignment has a target representation with zero generated constraint penalty" and adds that a finite continuous-slack encoding can violate that premise. This is the right fix — the objective-range exact-penalty argument (#173) relies on `ϵ` being the smallest *infeasible* penalty, which a nonzero feasible floor `p_F` breaks. The cross-link to the new section closes the loop.
- **Slack-ownership rules match `src/compiler/constraints.jl`.** `EqualTo` encodes its residual directly (no inequality slack); `LessThan`/`GreaterThan`/`Interval` sides generate one unless a sign-definite shortcut (`_is_nonnegative`/`_is_nonpositive`) or an always-feasible detection avoids it.
- **`Discretize(true)` is genuinely the default** (`src/attributes/compiler.jl:569`, `get(..., :discretize, true)`), and the integer- vs continuous-slack path description matches the `variable_ℤ!`/`variable_ℝ!` branch in the compiler.
- **The contrast math is consistent** with the executable analysis merged in #211 (`Δ = S/(2ⁿ−1)`, `p_F ≤ Δ²/4`, `ρ(p_I − p_F) > B`, and the `p_I ≤ p_F` no-repair case).
- **The metadata recipe is the real public API.** `ToQUBO.reformulation_metadata(JuMP.unsafe_backend(model))` and `metadata["slack_variables"]` are exactly the calls exercised by the #211 benchmark.
- **All `@ref` targets resolve unambiguously.** The new `### Slack Resolution and Penalty Contrast` heading occurs once and is referenced from both the booklet and settings; `Continuous Variable Resolution`, `Representation Error`, `Changing Penalty Values`, `Constraint Reformulation`, and the attribute docstrings (`Discretize`, `ConstraintPenaltyScale`, `PenaltyScale`, `PenaltyOffset`, `SlackVariableEncodingBits/ATol`) all exist. The external report blob links resolve on `main` (the #211 report is present at the base tip).
- **Issue linkage is accurate.** `Refs #208` (not `Closes` — #208 stays open for the design decision), and the PR correctly declines to present resolution-aware scaling as the #205 remedy, consistent with the #211 finding.

## Blocking

None.

## Nonblocking

None.

## Questions

None.

## Tests run

- `JULIA_LOAD_PATH="@:test:@stdlib" julia --project=. -e 'using Test; include("test/unit/slack_resolution_docs.jl"); test_slack_resolution_docs()'` → **13/13 passed** at head `40b7d10`.
- Non-vacuity: ran the same 13 assertions against the merge-base (`1177519`) docs content → **0 passed, 13 failed**. The guard genuinely fails without this change.
- Confirmed wiring: `runtests.jl` → `test_unit()` → `test_slack_resolution_docs()`.
- Working tree clean afterward; no artifacts.

## Merge-readiness

- Changed-file set matches `gh pr diff --name-only` (4 files, +140/−13).
- All required status checks green for head `40b7d10` (Julia 1 / 1.10 × ubuntu / windows), plus `build` and `documenter/deploy` — the authoritative Documenter build, which would fail on a broken `@ref`.
- Branch protection: `required_approving_review_count: 0`, `required_conversation_resolution: true`. This body-only review creates no threads, so it does not change merge state (`CLEAN`).

No blocking findings; the PR accurately documents the slack-resolution/penalty-contrast boundary and correctly qualifies the exact-penalty guarantee. This account cannot cast a formal approval, but branch protection does not require one.

PR #231, opened 2026-08-01
Add literature penalty-weight heuristics (UB-positive, max-coefficient, VLM, MOMC, MOC)

bernalde:
Closes #222

Part of the MQT QAO parity epic: JuliaQUBO/QUBO.jl#69

## Summary

Adds the five literature penalty-weight heuristics named in #222 as `AutomaticPenaltyPolicy` subtypes, selectable via `Attributes.PenaltyPolicy()`:

- `UBPositivePenalty` — sum of non-constant objective coefficients (Ayodele 2022, Eq. 12); valid only for nonnegative coefficients, guarded.
- `MaxCoefficientPenalty` — largest non-constant objective coefficient (Eq. 13, after Lucas 2014).
- `VLMPenalty` — largest one-flip objective change (Verma & Lewis 2022; Eqs. 14–15).
- `MOMCPenalty` — `max(1, VLM/γ)` with `γ` the smallest strictly positive one-flip change of the penalty function (Eqs. 16–18), **per penalty function**.
- `MOCPenalty` — `max(1, max |one-flip objective/penalty ratio|)` over positive penalty flips (Eq. 19), **per penalty function**.

Each bound composes exactly like the default policy (`ρ = s·σ·(λ+β)/ϵ`), applies uniformly to all three penalty buckets (constraints, variable encodings, slack encodings), and falls back per-coefficient to `LegacyPenalty` via the existing `_fallback!` machinery — with a specific reason string recorded in `PenaltyPolicyMetadata` — whenever its validity conditions fail (negative coefficients under UB⁺, empty objective, no positive penalty flip, non-finite ϵ).

## Design notes

- **Formulas follow the literature (Ayodele, arXiv:2206.11040), not MQT QAO's implementation.** Their code deviates from the paper it cites in ways we deliberately do not replicate: the constant offset added into MQC, the missing strictly-positive filter in MOMC's γ (their `min` admits negative flip values, collapsing λ to 1), and the missing absolute value in MOC's ratios. Each is a correctness deviation, verified against `mqt-qao/src/mqt/qao/problem.py:154-402`.
- One-flip values generalize Ayodele's quadratic row-sum form to arbitrary degree by crediting each monomial to every variable it contains — the correct flip bound for the PBF IR (and what makes the policies degree-agnostic).
- MOMC/MOC are per-penalty-function (per constraint/variable/slack), per the issue's acceptance criteria, rather than Ayodele's single aggregated constraint matrix; documented in the docstrings and manual.
[...]

PR #228, opened 2026-08-01
Surface feasibility in solve results (PrimalStatus, source objective value, auto-report)

bernalde:
Closes #223

Part of the MQT QAO parity epic: JuliaQUBO/QUBO.jl#69

## Summary

Samplers solve the unconstrained target model, so QUBODrivers unconditionally reports `FEASIBLE_POINT` for every in-range result (`src/library/sampler/wrappers/moi.jl:30-40`), and `MOI.ObjectiveValue` reports the penalized target energy. This PR wires ToQUBO's existing feasibility machinery (#138) into the solve path so results surface source-model semantics:

- **`MOI.PrimalStatus` feasibility check** — new `Attributes.PrimalFeasibilityCheck()` (default **on**): when the sampler reports `FEASIBLE_POINT`, ToQUBO re-checks the decoded sample against source constraints (lazily, cached per result index) and reports `INFEASIBLE_POINT` on violation. Turning the attribute off restores the previous behavior exactly.
- **Penalty-free source objective** — new `Attributes.SourceObjectiveValue(result_index)` plus exported `source_objective_value(model; result)`: evaluates the *source* objective function on the projected source-variable values. `MOI.ObjectiveValue` semantics are deliberately unchanged (the issue offered either option; silently changing a core MOI attribute's meaning seemed riskier than a dedicated, discoverable attribute).
- **Automatic feasibility report** — new `Attributes.AutoFeasibilityReport()` (default off): after each solve, computes the full `FeasibilityReport`, caches it, and warns (respecting `Attributes.Warnings()`) when any sample violates a source constraint. Default-argument `feasibility_report(model)` calls now return the per-solve cached report; caches are invalidated in `MOI.optimize!` and `Compiler.reset!`.

## Design notes

- Default-**on** for `PrimalFeasibilityCheck` is a deliberate behavior change: reporting `FEASIBLE_POINT` for constraint-violating samples is the misleading default the issue targets. Existing suite assertions only check result 1 of well-penalized models and are unaffected.
- The feasibility tolerance is the analysis default (`1e-6`), now shared via `_FEASIBILITY_ATOL` so `violations`/`is_feasible`/`feasibility_report` and the status check agree.
- `MOI.DualStatus` handling is untouched (split out of the former `Union` method).

## Acceptance criteria (from #223)

- [x] Infeasible samples report `INFEASIBLE_POINT`; feasible ones unchanged (`test_feasibility_primal_status`)
[...]

PR #232, opened 2026-08-02
Add value-set variable encoding input surface (arbitrary finite sets and non-uniform spacing)

### LINKED ISSUE #221: Add value-set variable encoding input surface (arbitrary finite sets and non-uniform spacing)
*This issue merges two verified gap rows — "discrete variables with arbitrary finite value sets" and "non-uniform (geometric/logarithmic) point spacing for continuous variables" — because both reduce to the same input-surface work and would ship as one PR.*

## Motivation
MQT QAO (arXiv:2406.12840) exposes `add_discrete_variable(name, [values])` accepting any float list, plus `distribution='geometric'/'logarithmic'` sampling for continuous variables (mqt-qao `src/mqt/qao/variables.py:171-244, 246-364, 862-907`). Their implementation is weaker than ours under the hood — one-hot dictionary encoding only (comment at `variables.py:902`), and the logarithmic mode misuses `np.logspace` (`variables.py:972`, min/max treated as base-10 exponents) — but we currently have no user-facing way to declare such variables at all.

## Current state
The entire stack below the compiler already handles arbitrary finite sets: `encode(var, OneHot/DomainWall, γ)` at `src/encoding/variables/set/one_hot.jl:27-46` and `set/domain_wall.jl:20-39`, and the virtual-layer entry point `Encoding.encode!(model, x, e, γ::AbstractVector)` at `src/virtual/encoding.jl:48-65`. But the compiler scans only ZeroOne/Integer/Semiinteger/Semicontinuous/Interval/LT/GT (`src/compiler/variables.jl:12-90`), continuous set-encoding paths always build an even grid `Γ = range(a, b; length = p)` (`one_hot.jl:100-108`, `domain_wall.jl:89-99`), and no attribute in `src/attributes/compiler.jl` carries a value list.

## Proposed change
- Add a per-variable compiler attribute (e.g. `Attributes.VariableEncodingSet <: CompilerVariableAttribute`), pattern-copying `VariableEncodingBits` at `src/attributes/compiler.jl:1032`.
- Branch in `variables!`/`variable_ℤ!` (`src/compiler/variables.jl:99-116, 240-254`) to call the γ-vector `encode!` when the attribute is set and the method is a `SetVariableEncodingMethod`.
- Accept an explicit point vector (covering geometric/log spacing generated by the user or a helper) for continuous variables, routed to `encode(var, e, γ)`.
- Interval-only methods (Binary/Unary/Arithmetic/Bounded) must reject value sets with a clean error. No `Quadratize` needed (one-hot penalty is quadratic); the variable penalty χ registers via `model.h` automatically.

## Acceptance criteria
- [ ] A variable with value set e.g. `[-1.0, 1.0, 3.0]` compiles via OneHot and DomainWall and decodes to a member of the set on every sample.
- [ ] Interval-only encoding methods raise a descriptive error when a value set is present.
- [ ] Non-uniformly spaced point vectors for continuous variables round-trip through encode/decode.
- [ ] Tests cover both set encodings and the error path; docs section added to `docs/src/manual/4-settings.md`.

Part of the MQT QAO parity epic: https://github.com/JuliaQUBO/QUBO.jl/issues/69

bernalde:
Closes #221

Part of the MQT QAO parity epic: JuliaQUBO/QUBO.jl#69

## Summary

Adds the missing input surface for variables over explicit finite value sets: a new per-variable compiler attribute `Attributes.VariableEncodingSet` routed by the compiler to the pre-existing γ-vector set encoders (`Encoding.OneHot`, `Encoding.DomainWall`).

```julia
set_attribute(x, Attributes.VariableEncodingMethod(), Encoding.OneHot())
set_attribute(x, Attributes.VariableEncodingSet(), [-1.0, 1.0, 3.0])
```

- **Attribute** (`src/attributes/compiler.jl`): `VariableEncodingSet <: CompilerVariableAttribute`, pattern-copied from `VariableEncodingBits` (per-variable storage, `nothing`-deletion, `variable_encoding_set` helper), values stored as `Vector{T}`.
- **Compiler** (`src/compiler/variables.jl`): `variable_ℤ!`/`variable_ℝ!` branch to a new `variable_set!` before the bounds checks, so a set-valued variable needs no bounds (the set is the domain). Validation with descriptive `CompilationError`s: non-set encoding methods (Binary/Unary/Arithmetic/Bounded) rejected; empty, non-finite, or duplicate sets rejected; non-integer entries rejected for `MOI.Integer` variables; entries outside explicit bounds rejected. Binary and semi-integer/semi-continuous variables reject the attribute outright instead of silently ignoring it.
- **Non-uniform continuous grids**: the same attribute accepts an arbitrary point vector (e.g. `exp10.(range(-1, 2; length = 4))`), covering MQT QAO's geometric/logarithmic distributions — generated by the user, so any spacing works, and without MQT's `np.logspace` misuse (their `variables.py:972` treats min/max as base-10 exponents).
- **Docs**: new "Value-Set Variable Encoding" section in the settings manual with a solvable example and the attribute docstring.

## Design notes

- **No auto-selection of a set encoding**: a value set with an interval method errors rather than silently switching the method, keeping `VariableEncodingMethod` truthful (and matching the issue's error acceptance criterion). The error names OneHot/DomainWall as the fix.
- The virtual layer, decode path, and one-hot/domain-wall penalty registration (`model.h`) needed **no changes** — the γ-vector `Encoding.encode!` entry point existed since the set encoders landed; this PR is purely the input surface, as the issue's verified analysis predicted.
- Encoding-violating samples (e.g. all-zero one-hot state) decode outside the set by construction in any one-hot scheme; the solve test asserts the acceptance criterion on encoding-feasible samples (exactly one hot bit) and that all three set members appear.

## Acceptance criteria (from #221)

- [x] `[-1.0, 1.0, 3.0]` compiles via OneHot and DomainWall and decodes to set members (`test_compiler_variables_value_set_onehot`/`_domain_wall`, exact expansions + wall-state decode checks)
- [x] Interval-only methods raise a descriptive error (`test_compiler_variables_value_set_errors`)
- [x] Non-uniform continuous point vectors round-trip (`test_compiler_variables_value_set_continuous`, log-spaced grid)
- [x] Tests cover both set encodings and error paths; docs section added to `docs/src/manual/4-settings.md`

## Testing

- `julia --project -e 'using Pkg; Pkg.test()'` (Julia 1.10.11): full suite including five new test functions (OneHot/DomainWall/continuous expansions, seven error paths, end-to-end JuMP+ExactSampler solve with optimum −1 over the set).

## Branch Hygiene

- Base: `main` (default), branched from `origin/main` at `ef6fa02` (post-#231 merge).
- Open sibling PRs #229/#230 are Dependabot workflow-only bumps; no overlap.

### DIFF
--- a/docs/src/manual/4-settings.md
+++ b/docs/src/manual/4-settings.md
@@ -289,8 +289,11 @@ require integer-valued entries. Interval encoding methods (`Encoding.Binary`,
 `Encoding.Unary`, `Encoding.Arithmetic`, `Encoding.Bounded`) reject value sets
 with a compilation error — choose a set encoding for the variable instead.
 For continuous variables, pass the exact grid you want (for example
-`exp10.(range(-1, 2; length = 4))` for logarithmic spacing); decoded solutions
-always land on a grid point.
+`exp10.(range(-1, 2; length = 4))` for logarithmic spacing). Encoding-feasible
+decoded states always land on a grid point; samples that violate the set
+encoding's penalty (for example, an all-zero or multi-hot one-hot state) can
+decode off-grid, so inspect sample feasibility or the penalty value as usual
+when working with raw sample sets.
 
 ```@docs
 ToQUBO.Attributes.VariableEncodingSet
--- a/docs/src/manual/4-settings.md
+++ b/docs/src/manual/4-settings.md
@@ -256,6 +256,46 @@ Very small `ϵ` values or broad objective ranges can produce large penalties.
 Inspect the policy metadata before overriding the policy or pinning explicit
 penalties.
 
+### Value-Set Variable Encoding
+
+Variables restricted to an explicit finite set of values — discrete component
+choices, tariff levels, or a non-uniform (for example, logarithmically spaced)
+grid for a continuous quantity — can be declared directly with
+[`ToQUBO.Attributes.VariableEncodingSet`](@ref) together with a set encoding
+method (`Encoding.OneHot` or `Encoding.DomainWall`):
+
+```@example value-set-encoding
+using JuMP
+using QUBODrivers
+using ToQUBO
+using ToQUBO: Attributes, Encoding
+
+model = Model(() -> ToQUBO.Optimizer(ExactSampler.Optimizer))
+
+@variable(model, x, Int)
+@objective(model, Min, x)
+
+set_attribute(x, Attributes.VariableEncodingMethod(), Encoding.OneHot())
+set_attribute(x, Attributes.VariableEncodingSet(), [-1.0, 1.0, 3.0])
+
+optimize!(model)
+
+value(x)
+```
+
+The set replaces the bounds-derived domain, so bounds are optional; when
+explicit bounds are present, every entry must respect them. Integer variables
+require integer-valued entries. Interval encoding methods (`Encoding.Binary`,
+`Encoding.Unary`, `Encoding.Arithmetic`, `Encoding.Bounded`) reject value sets
+with a compilation error — choose a set encoding for the variable instead.
+For continuous variables, pass the exact grid you want (for example
+`exp10.(range(-1, 2; length = 4))` for logarithmic spacing); decoded solutions
+always land on a grid point.
+
+```@docs
+ToQUBO.Attributes.VariableEncodingSet
+```
+
 ### Continuous Variable Resolution
 
 Bounded continuous variables need a finite binary representation before the
--- a/src/compiler/variables.jl
+++ b/src/compiler/variables.jl
@@ -269,6 +269,13 @@ function _sos1_domain_wall_variables!(
             xi -> xi in binary_variables && !haskey(model.source, xi) && counts[xi] == 1,
             x.variables,
         )
+            # This fast path encodes binary variables before the per-variable
+            # loop, so it must apply the same value-set rejection the loop
+            # would reach through `variable_𝔹!`.
+            for xi in x.variables
+                _reject_value_set!(model, xi, "binary")
+            end
+
             _encode_sos1_domain_wall!(model, ci, x.variables)
         end
     end
--- a/src/compiler/variables.jl
+++ b/src/compiler/variables.jl
@@ -119,7 +119,77 @@ function variables!(model::Virtual.Model{T}, ::AbstractArchitecture) where {T}
 end
 
 function variable_𝔹!(model::Virtual.Model{T}, i::Union{VI,CI}) where {T}
+    _reject_value_set!(model, i, "binary")
+
     return Encoding.encode!(model, i, Encoding.Mirror{T}())
+end
+
+function _reject_value_set!(model::Virtual.Model, i::Union{VI,CI}, kind::String)
+    if i isa VI && !isnothing(Attributes.variable_encoding_set(model, i))
+        compilation_error!(
+            model,
+            "Value sets are not supported for $(kind) variables; found one on variable '$(i)'";
+            status = "Value set on $(kind) variable",
+        )
+    end
+
+    return nothing
+end
+
+# Encodes a variable over an explicit finite value set through a set encoding
+# method (one-hot, domain-wall). The set replaces the bounds-derived domain;
+# explicit bounds, when present, must contain every entry.
+function variable_set!(
+    model::Virtual.Model{T},
+    vi::VI,
+    γ::Vector{T},
+    (a, b)::Tuple{A,B};
+    integer::Bool,
+) where {T,A<:Union{T,Nothing},B<:Union{T,Nothing}}
+    e = Attributes.variable_encoding_method(model, vi)
+
+    if !(e isa Encoding.SetVariableEncodingMethod)
+        compilation_error!(
+            model,
+            "Variable encoding method '$(nameof(typeof(e)))' does not support value sets; " *
+            "choose a set encoding such as 'Encoding.OneHot' or 'Encoding.DomainWall' " *
+            "for variable '$(vi)'";
+            status = "Value set requires a set encoding method",
+        )
+    elseif isempty(γ)
+        compilation_error!(
+            model,
+            "Value set for variable '$(vi)' is empty";
+            status = "Empty value set",
+        )
+    elseif !all(isfinite, γ)
+        compilation_error!(
+            model,
+            "Value set for variable '$(vi)' contains non-finite entries";
+            status = "Non-finite value set",
+        )
+    elseif !allunique(γ)
+        compilation_error!(
+            model,
+            "Value set for variable '$(vi)' contains duplicate entries";
+            status = "Duplicate value set entries",
+        )
+    elseif integer && !all(isinteger, γ)
+        compilation_error!(
+            model,
+            "Value set for integer variable '$(vi)' must contain only integer values";
+            status = "Non-integer value set entries",
+        )
+    elseif (!isnothing(a) && any(v -> v < a, γ)) || (!isnothing(b) && any(v -> v > b, γ))
+        compilation_error!(
+            model,
+            "Value set for variable '$(vi)' has entries outside its declared bounds " *
+            "[$(isnothing(a) ? "-∞" : a), $(isnothing(b) ? "+∞" : b)]";
+            status = "Value set outside variable bounds",
+        )
+    end
+
+    return Encoding.encode!(model, vi, e, γ)
 end
 
 function _sos1_variable_counts(model::Virtual.Model{T}) where {T}
@@ -211,6 +281,8 @@ function variable_semiinteger!(
     vi::VI,
     S::Tuple{T,T},
 ) where {T}
+    _reject_value_set!(model, vi, "semi-integer")
+
     e = Attributes.variable_encoding_method(model, vi)
 
     MOI.set(model, Attributes.Quadratize(), true)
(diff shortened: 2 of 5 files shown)


bernalde (commented):
Maintainer review of head `fa5fef4` (author account — COMMENT review; cannot serve as formal approval).

## Scope and design

Resolves #221 as pure input surface over the pre-existing γ-vector set encoders, which is the right shape: no second encoding pathway, no virtual-layer or decode changes, and the `model.h` penalty registration comes free. Design decisions reviewed and endorsed:

- No auto-selection of a set encoding: a value set under an interval method errors with a message naming `Encoding.OneHot`/`Encoding.DomainWall`, keeping `VariableEncodingMethod` truthful and matching the issue's acceptance criterion.
- The set replaces the bounds-derived domain (bounds optional; contradicting explicit bounds is an error, not a silent intersection). Binary and semi-variables reject the attribute rather than silently ignoring it — the silent-ignore trap is closed.
- The solve test honestly handles the one inherent limitation: encoding-violating one-hot states decode outside the set in any such scheme, so the per-sample criterion is asserted over encoding-feasible samples (exactly one hot bit), with all three set members required to appear.

## Findings

**Nonblocking: no dedicated attribute round-trip testset** 
Repo convention gives per-variable/compiler attributes explicit round-trip testsets in `test/unit/attributes/attributes.jl` (e.g. `DefaultVariableEncodingMethod` at :487-496, the `PrimalFeasibilityCheck`/`AutoFeasibilityReport` sets added in #228). `VariableEncodingSet` is exercised only through compiler tests. Add a testset asserting: `MOI.supports`, default `nothing`, set-then-get returning the converted `Vector{Float64}` (e.g. setting `[-1, 1, 3]::Vector{Int}` reads back as `Vector{Float64}`), and `nothing`-deletion restoring `nothing`.

**Nonblocking: `Encoding.Bounded` and continuous DomainWall missing from the test matrix** 
The interval-method error test loops over default(Binary)/Unary/Arithmetic but not `Encoding.Bounded(...)` (the wrapping method — same rejection path, currently unasserted), and the set-encoding feature tests cover integer DomainWall + continuous OneHot but not continuous DomainWall. One loop entry and one small test close both diagonals of the matrix.

## Verification performed

- Full local suite at exactly this tree (Julia 1.10.11): green, including the five new test functions (exact one-hot/domain-wall/continuous expansions, seven error paths, end-to-end ExactSampler solve).
- Standalone smoke at implementation time: integer variable over `[-1, 1, 3]` compiles to the expected `ξ = -y₁ + y₂ + 3y₃` expansion with one-hot χ registered.
- CI green on head `fa5fef4`: Julia 1.10/1 × ubuntu/windows, docs build (zero missing-docstring warnings, matching base — the new attribute docstring renders in the settings manual `@docs` block), labeler.
- File-set match between local merge-base diff and `gh pr diff`; base `main` (`ef6fa02`, post-#231) unchanged since branch point, so PR-head checks equal merge-result checks. Open sibling PRs are Dependabot workflow-only bumps.
- Linked-issue check: all four acceptance criteria satisfied; `Closes #221` is correct.

## Summary

- Blocking: none.
- Nonblocking: 2 (attribute round-trip testset, Bounded/continuous-DomainWall matrix entries).
- Questions: none.
- Merge-readiness: CI green; this COMMENT review cannot formally approve.

bernalde:
Addressed both nonblocking findings from the [maintainer review](https://github.com/JuliaQUBO/ToQUBO.jl/pull/232#pullrequestreview-4837121081) in de89cb4:

- `attribute-round-trip` — added a `VariableEncodingSet` testset to `test/unit/attributes/attributes.jl` following the `DefaultVariableEncodingMethod` pattern: `MOI.supports`, default `nothing`, integer-vector input converting to `Vector{Float64}` on readback, helper accessor, and `nothing`-deletion.
- `matrix-gaps` — the interval-method rejection loop now includes `Encoding.Bounded(Encoding.Binary(), 2.0)`, and a new `test_compiler_variables_value_set_continuous_domain_wall` covers the continuous × DomainWall diagonal with exact wall-state decode assertions over a non-uniform grid.

Tests: full local suite (`Pkg.test()`, Julia 1.10.11) green at de89cb4, zero failures. CI on the new head is running.

No findings were declined.

src/compiler/variables.jl
```diff
@@ -119,9 +119,79 @@ function variables!(model::Virtual.Model{T}, ::AbstractArchitecture) where {T}
 end
 
 function variable_𝔹!(model::Virtual.Model{T}, i::Union{VI,CI}) where {T}
+    _reject_value_set!(model, i, "binary")
```
bernalde, line 122: Blocking: this rejection is bypassed for binary variables in the SOS1 DomainWall fast path. `variables!` calls `_sos1_domain_wall_variables!` before the per-variable loop; that helper inserts each eligible binary into `model.source`, so the loop skips it and never calls `variable_𝔹!`. I reproduced this at `de89cb4` with two `ZeroOne` variables in an SOS1 constraint and `VariableEncodingSet() = [-1.0, 3.0]` on one variable: compilation succeeds, the set remains stored, and the variable receives the ordinary 0/1 DomainWall expansion (`y₁ - y₂`). Please reject value sets before the SOS1 pre-pass (or exclude attributed variables from that fast path) and add a regression test asserting a compilation error for an eligible SOS1 binary carrying this attribute.

docs/src/manual/4-settings.md
```diff
@@ -256,6 +256,46 @@ Very small `ϵ` values or broad objective ranges can produce large penalties.
...
+```
+
+The set replaces the bounds-derived domain, so bounds are optional; when
+explicit bounds are present, every entry must respect them. Integer variables
+require integer-valued entries. Interval encoding methods (`Encoding.Binary`,
+`Encoding.Unary`, `Encoding.Arithmetic`, `Encoding.Bounded`) reject value sets
+with a compilation error — choose a set encoding for the variable instead.
+For continuous variables, pass the exact grid you want (for example
+`exp10.(range(-1, 2; length = 4))` for logarithmic spacing); decoded solutions
+always land on a grid point.
```
bernalde: Nonblocking: `always` overstates what the penalty-based encodings guarantee. The PR's own OneHot solve test correctly filters to encoding-feasible samples because an all-zero or multi-hot sample can decode outside the supplied set (for example, all-zero decodes to `0`, which is not in `[-1, 1, 3]`). Please qualify this as encoding-feasible decoded states landing on a grid point, and make the same qualification in the attribute docstring so users know to inspect feasibility/penalty status.

bernalde (commented):
Maintainer re-review of head `de89cb4` (previously reviewed at `fa5fef4`). The follow-up commit fully addresses both earlier nonblocking requests: `VariableEncodingSet` now has a dedicated attribute round-trip test, and the compiler matrix now covers `Encoding.Bounded` rejection plus continuous DomainWall.

I found one new blocking correctness issue and one nonblocking documentation issue, both left inline. In particular, an eligible SOS1 binary variable bypasses the new binary-variable rejection and silently compiles with ordinary 0/1 DomainWall semantics even when a different value set is stored. That contradicts the advertised contract and can silently encode the wrong domain.

Verification: the full exact-head Julia 1.10.11 suite passes (1927/1927), all current GitHub checks are green across Julia 1/1.10 on Ubuntu/Windows plus docs and labeler, and the five-file local diff matches GitHub. The linked issue scope and sibling-PR overlap were also checked.

Because this account is also the PR author, GitHub does not allow me to submit `REQUEST_CHANGES`; this is therefore a COMMENT review. I would not merge until the blocking SOS1 bypass is fixed and covered by a regression test.

bernalde:
Addressed the round-2 review (head `de89cb4`) in 1204862:

- **Blocking — SOS1 fast path bypasses the binary value-set rejection** (inline on `_sos1_domain_wall_variables!`): confirmed the bug first with your exact reproduction on the unfixed head — two `ZeroOne` variables in an SOS1 with `VariableEncodingSet() = [-1.0, 3.0]` compiled silently and produced the ordinary `y₁ - y₂` domain-wall expansion, ignoring the set. The fast path now applies `_reject_value_set!` to every constraint variable before `_encode_sos1_domain_wall!`, so it enforces the same contract `variable_𝔹!` would; the reproduction now throws `CompilationError: Value sets are not supported for binary variables...`. Regression test added (`test_compiler_variables_value_set_sos1_rejected`), which fails on the unfixed code and passes on this head.
- **Nonblocking — "always land on a grid point" overstates the guarantee** (inline on the manual): qualified in both places to encoding-feasible decoded states, with the off-grid caveat for penalty-violating samples (all-zero/multi-hot) and a pointer to inspect sample feasibility — the manual sentence and the `VariableEncodingSet` docstring now match the solve test's own filtering.

Tests: full local suite (`Pkg.test()`, Julia 1.10.11) green at 1204862; CI on the new head is running.

No findings were declined. Thread resolution is left to the human reviewer.

src/compiler/variables.jl
bernalde: Addressed in 1204862: confirmed your reproduction on the unfixed head (silent compile, `y₁ - y₂` expansion with the set ignored), then added `_reject_value_set!` over every constraint variable in the SOS1 fast path before `_encode_sos1_domain_wall!` — the same contract `variable_𝔹!` enforces. The reproduction now throws `CompilationError`, and the new `test_compiler_variables_value_set_sos1_rejected` regression fails on the unfixed code and passes at head.

docs/src/manual/4-settings.md
bernalde: Addressed in 1204862: both the manual sentence and the `VariableEncodingSet` docstring now say encoding-feasible decoded states land on a grid point, with the off-grid caveat for penalty-violating samples and a pointer to inspect sample feasibility — matching the solve test's encoding-feasible filtering.

Issue #233, opened 2026-08-02
Add subgradient (augmented-Lagrangian) penalty update strategy


bernalde:
## Motivation
The iterative penalty-refinement engine (JuliaQUBO/QUBO.jl#67, design record in [this comment](https://github.com/JuliaQUBO/QUBO.jl/issues/67#issuecomment-5159885555)) ships with multiplicative escalation for MQT QAO parity. Multiplicative escalation needs ρ→∞ for feasibility and collides with sampler coefficient dynamic-range limits within a few ×10 steps (Ohzeki, Sci. Rep. 10:3126, 2020, names this explicitly). The method of multipliers converges with **bounded** ρ (Powell's shifted-penalty view; Bertsekas 1982: linear rate at fixed finite ρ), and Yonaga–Miyama–Ohzeki (arXiv:2012.06119) run exactly this iteration on a D-Wave annealer with fixed ρ = 0.1, tracking best-feasible and best-energy samples separately.

## Proposed change
Building on the engine from #67's PR A:

- `Attributes.AugmentedLagrangianPenalty{T} <: ConstraintPenaltyMethod` with `multiplier::T` (λ, any sign, default 0) and `rho::T` (> 0): emits `λ·ℓ + ρ·ℓ²` for equalities and the sign-oriented composite for LessThan/GreaterThan, reusing `_unbalanced_penalty`'s emission (`src/compiler/constraints.jl:61-71`) but lifting `UnbalancedPenalty`'s strictly-positive-coefficients and inequality-only restrictions. Document as *iterated unbalanced penalization* for inequalities (the Rockafellar one-sided AL term is not QUBO-expressible slack-free).
- `SubgradientUpdate` strategy for `Attributes.PenaltyUpdateStrategy`: per iteration, from the best sample's **signed** residuals (`_constraint_measurements`; EqualTo residual is signed, inequalities positive-when-violated): equality `λ ← λ + η·r`; inequality `λ ← max(0, λ + η·r)`; default step η = ρ (classical MoM). Safeguard: escalate ρ ×τ only when `‖r_k‖ > γ·‖r_{k−1}‖` (Bertsekas γ ≈ 0.25, BCL-style), so escalations stay rare and coefficients bounded.
- MAX-sense handling via the compiler's σ convention (multipliers live in the minimization frame; tests must cover a MAX model).
- Interval constraints: the measured residual loses violation direction (`feasibility.jl:357-367`) — either split handling or document exclusion.

[...]

Issue #138, opened 2026-06-10
Post-sampling feasibility and constraint-violation reporting via reformulation metadata


bernalde:
Part of the benchmarking-readiness epic: JuliaQUBO/QUBO.jl#40 (WS5).

## Problem

After sampling a ToQUBO-reformulated model, there is **no built-in way to ask whether a returned sample satisfies the original constraints**, nor by how much it violates them. Solutions are projected back to original variables (`src/reformulation.jl`, `project_original_state`, v0.4) but feasibility checking is left entirely to the user. For the QOBLib benchmarking campaign this is a headline metric: for constrained classes, "best energy found" is meaningless without "…and the sample was feasible", and penalty-tuning (companion issue) can't be evaluated without violation measurements.

The v0.4 reformulation metadata already tracks original/auxiliary variables and supports state projection — the foundation exists; this issue builds the reporting layer on it.

## Proposed PR

1. **Core API** (suggest `src/analysis/feasibility.jl`):
   ```julia
   violations(model::JuMP.Model; result::Int = 1) -> Vector{ConstraintViolation}
   is_feasible(model; result, atol=1e-6) -> Bool
   feasibility_report(model) -> per-result summary (feasible count, max/total violation, by constraint type)
   ```
   where `ConstraintViolation` carries the constraint index, type, violation magnitude, and the projected original-variable values involved.
2. **Mechanics**: for each source constraint stored during compilation, evaluate the original function on the projected state (LHS value) and measure distance to the set (`MOI.Utilities.distance_to_set` covers the standard scalar sets — reuse it rather than reimplementing).
3. **Integer-projection caveat**: projected values from binary encodings are exact for integer encodings but approximate for continuous discretizations — report violations *after* rounding per the encoding's semantics, and expose the raw residual too.
4. **Per-sample iteration**: operate on any `result` index so the harness can compute feasibility *rates* over a whole `SampleSet`, not just the incumbent.
5. **Docs + example**: knapsack example in the manual: sample with `RandomSampler`, show the feasibility report, show how the rate changes with `PenaltyScale` (ties to the penalty parameterization issue).

Out of scope (file follow-ups if wanted): solution *repair* heuristics; feasibility-aware sample filtering inside QUBODrivers (`PostSampleCallback` already exists there as a hook — mention in docs).

## Checkpoints (anti-drift)

[...]

Issue #223, opened 2026-07-31
Surface feasibility in solve results (PrimalStatus, source objective value, auto-report)


bernalde:
## Motivation
MQT QAO (arXiv:2406.12840) auto-runs `check_constraint_optimal_solution` after every solve, and its Solution object exposes `check_constraint_all_solutions` and a valid-solution rate (`src/mqt/qao/solvers.py:1503-1530`; `docs/Solution.rst:27-41`) — albeit with operator-precedence bugs in its AND/OR checkers (`constraints.py:402,458,557,613`). Our checking machinery is sounder (numeric `MOIU.distance_to_set` residuals, `feasibility.jl:320-331`) but is not wired into the solve path, so practitioners get misleading statuses by default.

## Current state
Machinery exists: `violations`/`is_feasible`/`feasibility_report` (`src/analysis/feasibility.jl:69-117`) — and note the aggregate valid-sample rate ALREADY exists: `feasibility_report(model)` with `result=nothing` summarizes every result with `feasible_count`/`result_count` plus violation histograms (`feasibility.jl:48-54, 102-117, 255-277`). The confirmed gaps: (1) nothing runs automatically after `optimize!` (`src/wrapper.jl:27-44`); (2) `PrimalStatus` is unconditionally `FEASIBLE_POINT` (QUBODrivers `src/library/sampler/wrappers/moi.jl:29-39`; ToQUBO `src/attributes/solver.jl:124-138` just delegates); (3) `MOI.ObjectiveValue` returns the penalized QUBO energy, not the source objective (`src/attributes/solver.jl:150-156`).

## Proposed change
- Override `MOI.get(::Virtual.Model, ::MOI.PrimalStatus)` in `src/attributes/solver.jl` to consult the feasibility analysis for the given result index (cache per result; guard on `ResultCount > 0`).
- Add penalty-free source-objective reporting: evaluate `model.f` on the sample state (same mechanism as `_evaluate_pbf` in `src/reformulation.jl`) — either override `MOI.ObjectiveValue` (behind an attribute, to avoid silently changing semantics) or add a dedicated `Attributes.SourceObjectiveValue`.
- Optionally auto-run and store a `FeasibilityReport` after `optimize!` in `src/wrapper.jl:27-44`, gated by an attribute, so JuMP users get the valid rate without extra calls (`ext/ToQUBOJuMPExt.jl:13-24` wrappers already exist).

## Acceptance criteria
[...]

PR #234, opened 2026-08-02
Add iterative penalty refinement engine with multiplicative strategy

### LINKED ISSUE #67: Add iterative self-adaptive penalty refinement (solve, check, rescale violated penalties, re-solve)
## Motivation
MQT QAO's flagship capability (arXiv:2406.12840): every solver loops up to `max_lambda_update=5` times, updating ONLY violated constraints' lambdas via "sequential penalty increase" (x10), "scaled sequential", or "binary search", then re-quadratizes and re-solves (`src/mqt/qao/problem.py:104-152, 404-443`; `solvers.py:154-204`, default at `:1035+`; increases only, never decreases). Mis-tuned penalties producing infeasible samples is the single most practical pain point for practitioners.

## Current state
Verified missing ecosystem-wide: `penalties!` runs once during compile with no feedback (ToQUBO `src/compiler/penalties.jl:199-291`); `MOI.optimize!` is a single `compile!` then delegate (ToQUBO `src/wrapper.jl:27-44`); grep for penalt/lambda/adaptive/iterat over QUBODrivers.jl and QUBO.jl src: nothing. All building blocks exist: `Analysis.violations` (`src/analysis/feasibility.jl:69-78`), per-constraint `ConstraintPenaltyScale` (`src/attributes/compiler.jl:1457-1481`), `Compiler.reset!`/`compile!` (`src/compiler/compiler.jl:78, 37/45`), and manual-sweep docs (`docs/src/manual/4-settings.md:131-157`).

## Proposed change
Add an attribute-gated loop inside `MOI.optimize!(model::Optimizer)` in ToQUBO `src/wrapper.jl:27-44` (or a meta-optimizer in QUBO.jl `src/wrapper.jl` wrapping any QUBODrivers sampler). Per iteration: (1) run sampler; (2) `Analysis.violations(model)` on the best result; (3) for each violated source ci, multiply `Attributes.ConstraintPenaltyScale` by a factor (x10, or binary-search style `sqrt(lambda*upper)`); (4) `Compiler.reset!` + `Compiler.compile!` and re-delegate. New compiler attributes (`MaxPenaltyUpdates`, `PenaltyUpdateStrategy`) follow the `ConstraintPenaltyScale` pattern (`src/attributes/compiler.jl:1445-1481`).

Pitfalls: explicit `ConstraintEncodingPenaltyHint` bypasses scale/offset (`penalties.jl:207-239`) — hinted constraints need direct hint multiplication; variable/slack encoding penalties (`model.h`/`model.s` buckets, `penalties.jl:242-284`) are additional infeasibility sources MQT does not touch but we could; guard on `ResultCount > 0` before the feasibility check (`feasibility.jl:200-208`). Per-iteration full recompile is acceptable (matches MQT).

## Acceptance criteria
- [ ] Opt-in via attributes; default behavior unchanged (single solve).
- [ ] A deliberately under-penalized test model reaches feasibility within the iteration budget.
- [ ] Hinted constraints are also escalated.
- [ ] Loop terminates on feasibility, budget exhaustion, or `ResultCount == 0`; iteration count/final scales are queryable.

Part of the MQT QAO parity epic: https://github.com/JuliaQUBO/QUBO.jl/issues/69

bernalde:
Closes JuliaQUBO/QUBO.jl#67
Refs #233 (subgradient/augmented-Lagrangian strategy, stacked follow-up)

Part of the MQT QAO parity epic: JuliaQUBO/QUBO.jl#69. Design record from the exploration stage: https://github.com/JuliaQUBO/QUBO.jl/issues/67#issuecomment-5159885555

## Summary

Adds the iterative penalty-refinement engine inside ToQUBO's `MOI.optimize!`, with a pluggable strategy interface and the MQT-parity multiplicative strategy:

- **Engine** (`src/refinement.jl`, hook in `src/wrapper.jl`): when `Attributes.MaxPenaltyUpdates() > 0`, after each solve the best sample's source-constraint residuals are measured (`_constraint_measurements`, the #138/#223 machinery); while any constraint is violated and the budget remains, penalties are updated, the model is fully recompiled (user attributes survive `Compiler.reset!`), and the sampler runs again. Stops on feasibility, budget exhaustion, empty result set, or when no violated constraint has a penalty coefficient to update (variable-encoding-driven violations have no per-constraint knob).
- **Strategy interface**: `Attributes.PenaltyUpdate` abstract type + `Attributes.PenaltyUpdateStrategy` attribute (mirroring the `AutomaticPenaltyPolicy` pattern), so #233's subgradient strategy plugs in without engine changes.
- **`MultiplicativeUpdate(factor = 10.0)`** — MQT QAO parity: violated constraints escalate ×factor, branching on the hint-vs-scale precedence: hinted constraints multiply the hint (sign preserved, so MAX-sense negative penalties escalate correctly — the issue's "hinted constraints are also escalated" criterion); inferred constraints multiply `ConstraintPenaltyScale` (sign-neutral; also escalates the constraint's slack-encoding penalty).
- **Readback**: `Attributes.PenaltyUpdateCount()` (`is_set_by_optimize`), cleared by `Compiler.reset!`; final coefficients queryable through the existing hint/scale/applied-penalty attributes.
- **Docs**: "Iterative Penalty Refinement" section in the settings manual with a solvable example, the coefficient-range caveat, and a pointer to #233.

## Design notes (from the recorded study)

- Placement in ToQUBO rather than a QUBO.jl meta-optimizer: QUBO.jl has no MOI machinery, and in-`optimize!` iteration keeps JuMP dirty-model semantics out of the loop while every existing accessor reports the final iterate.
- Full recompile per iteration is deliberate (matches MQT; correctness first — `UnbalancedPenalty`-style coefficients are baked into `model.g` at `constraints!` time, so an incremental `penalties!`+`build!` path would silently miss them; noted as a future optimization).
- MQT-contract fidelity: their "scaled sequential" and "binary search" mechanisms are geometric climbs toward a memoized 100×λ₀ ceiling with an integer-`round()` absorbing-zero bug and solver-instance state that leaks across solves — we implement the honest multiplicative contract and deliberately none of those behaviors.

## Acceptance criteria (from QUBO.jl#67)

- [x] Opt-in via attributes; default behavior unchanged (`test_refinement_default_off` asserts the under-penalized solve stays infeasible with untouched hints)
- [x] A deliberately under-penalized model reaches feasibility within budget (`test_refinement_hinted_escalation`, `test_refinement_scale_escalation` — both senses, exact iteration counts 2 and 6 hand-derived from the threshold `scale < 3/7`)
- [x] Hinted constraints are also escalated (hint −0.1 → −10.0 asserted)
- [x] Terminates on feasibility/budget/`ResultCount == 0`; iteration count and final coefficients queryable (`PenaltyUpdateCount`, budget-exhaustion test)

## Testing

- `julia --project -e 'using Pkg; Pkg.test()'` (Julia 1.10.11): full suite green, including the refinement tests in `test/unit/refinement.jl` (own file per branch-hygiene convention) covering escalation traces for hinted/scaled/slack paths, budget exhaustion, the opt-in default, constructor validation, and the `ResultCount == 0` stop condition via a counting probe sampler.

## Branch Hygiene

- Base: `main` (default), branched from `origin/main` at `4fb1532` (post-#232 merge).
- #235 (subgradient/augmented-Lagrangian strategy, closing #233) is stacked on this branch: merge this PR first, after which #235 retargets to `main`. No other open PR touches these files.


bernalde (commented):
Maintainer review of head `d483062` (author account — COMMENT review; cannot serve as formal approval).

## Scope and design

Implements the engine slice of the QUBO.jl#67 design record: attribute-gated loop inside `optimize!`, pluggable `PenaltyUpdate` strategy mirroring the `AutomaticPenaltyPolicy` pattern, and the MQT-parity multiplicative strategy. Reviewed and endorsed:

- Placement and full-recompile-per-iteration match the recorded design (user attributes survive `reset!`; an incremental `penalties!`+`build!` path would silently miss method-baked coefficients — correctly deferred).
- The hint-vs-scale branch handles the issue's "hinted constraints are also escalated" criterion, with sign preservation for MAX-sense hints; scales stay sign-neutral, and the non-updatable-violation guard (variable-encoding-driven infeasibility) prevents futile recompiles.
- Ordering is right: refinement runs before `_auto_feasibility_report!`, so #223's reporting reflects the final iterate; per-iteration `reset!` clears the feasibility caches.
- Deliberate non-replication of MQT's rounding/instance-state bugs is documented in the PR rather than silently diverging.

## Findings

**Nonblocking: slack-escalation claim untested** 
The scale path's comment (and PR body) state that escalating `ConstraintPenaltyScale` also escalates the constraint's slack-encoding penalty, but no test asserts it: both refinement fixtures have integer-slack constraints whose infeasibility is driven by the constraint penalty alone. Add a fixture asserting `SlackVariableEncodingPenalty` grows across iterations (the `slack_encoding = Encoding.OneHot()` recipe from the penalty tests gives a slack bucket directly), or drop the claim from the comment.

**Nonblocking: only the last iteration's `CompilationTime` survives** 
Each iteration overwrites `CompilationTime`, so after refinement the attribute silently reports the final recompile only — a user summing compilation + solve time gets a misleading total for a 5-iteration run. Either accumulate across the loop's recompiles or state the last-iteration semantics in the manual section (one sentence suffices; the iteration count is queryable so the reader can infer).

## Verification performed

- Full local suite green at exactly this tree (Julia 1.10.11), including seven refinement tests with hand-derived iteration counts (2 for ×10, 6 for ×2 from the `scale < 3/7` threshold; hint trace −0.1 → −10.0).
- CI green on head `d483062` (Julia 1.10/1 × ubuntu/windows, docs incl. the new solvable manual example, labeler); docs build has zero missing-docstring warnings (all five new docstrings render).
- File-set match local-vs-GitHub; base `main` (`4fb1532`) unchanged since branch point. Sibling PRs: Dependabot workflow-only, plus stacked #235 which builds on this branch (reviewed separately).
- Cross-repo `Closes JuliaQUBO/QUBO.jl#67` verified against that issue's four acceptance criteria — all satisfied at this head.

## Summary

- Blocking: none.
- Nonblocking: 2 (slack-escalation test, `CompilationTime` semantics).
- Questions: none.
- Merge-readiness: CI green, `mergeStateStatus: CLEAN`. This COMMENT review cannot formally approve.

bernalde:
Addressed both nonblocking findings from the [maintainer review](https://github.com/JuliaQUBO/ToQUBO.jl/pull/234#pullrequestreview-4839535788) in 403fd55:

- `slack-escalation-test` — added `test_refinement_slack_escalation`: the scaled fixture with a one-hot slack encoding asserts the slack-bucket coefficient lands at the hand-derived η = −7.0 after the scale escalates 0.01 → 1.0, proving the scale path drives the slack-encoding penalty as claimed.
- `compilation-time-semantics` — documented in the manual's refinement section: each iteration fully recompiles and `CompilationTime` reports the **last** recompile only, with total overhead scaling with the queryable iteration count.

Tests: full local suite (`Pkg.test()`, Julia 1.10.11) green at 403fd55. The fix is also merged forward into the stacked #235 branch (15bd3ff) so the stack stays consistent; CI is running on both heads.

No findings were declined.

src/attributes/compiler.jl
```diff
@@ -269,10 +269,56 @@ Reference: M. Ayodele, [arXiv:2206.11040](https://arxiv.org/abs/2206.11040).
...
+subgradient (augmented-Lagrangian) strategy tracked in ToQUBO#233.
+"""
+struct MultiplicativeUpdate <: PenaltyUpdate
+    factor::Float64
+
+    function MultiplicativeUpdate(factor::Real = 10.0)
+        isfinite(factor) && factor > 1 ||
+            throw(ArgumentError("Penalty update factor must be finite and greater than one"))
+
+        return new(Float64(factor))
```
bernalde: Blocking: validate the value after conversion to `Float64`. The `Real` input can be finite even when storage conversion overflows: at this head, `Attributes.MultiplicativeUpdate(big"1e400")` succeeds with `factor == Inf`, so the constructor violates its documented finite-factor contract and can feed non-finite penalty updates into compilation. Convert first, validate the converted value, then store it; add a regression case for a finite `BigFloat` that overflows `Float64`.

test/unit/refinement.jl
```diff
@@ -0,0 +1,228 @@
...
+function test_refinement()
+    @testset "→ Penalty Refinement" verbose = true begin
+        test_refinement_attributes()
+        test_refinement_hinted_escalation()
+        test_refinement_scale_escalation()
+        test_refinement_default_off()
+        test_refinement_already_feasible()
+        test_refinement_budget_exhaustion()
+        test_refinement_slack_escalation()
+        test_refinement_smaller_factor()
```
bernalde, line 331: Blocking: add coverage for the linked issue's `ResultCount == 0` stop condition. No refinement test supplies an empty result set; I deleted the early-return guard in `_best_sample_violations`, reran `Pkg.test()`, and all 1,975 tests still passed, confirming this contract can regress unnoticed. Add a no-result mock sampler fixture with refinement enabled and assert optimization returns without another update/solve and `PenaltyUpdateCount() == 0` (or exercise a first-result then zero-result sequence and assert the exact call/update count).

bernalde (commented):
Re-review of head `403fd556` after my prior author-account review of `d483062`.

The two prior nonblocking findings are resolved at this head: the manual now states that `CompilationTime` reports only the last recompile, and the slack path has an integration assertion. I also put the slack-scale propagation defect back in a disposable copy; the existing compiler scale/offset test failed, confirming the combined suite guards that behavior.

## Finding

**Nonblocking: refresh the PR's volatile test/branch metadata** 

The Testing section still says there are seven new refinement test functions, while this head defines eight (and the requested zero-result coverage will change that again). The Branch Hygiene section also says open sibling PRs are Dependabot-only, but the live set is now stacked #235 plus unrelated docs PR #236. Update those statements after the test fix, or remove the brittle count/open-PR inventory and retain only the durable coordination note that #235 is stacked on this branch.

## Verification performed

- Full local suite at the exact PR head: `julia --project=/tmp/toqubo-pr234.IwCEzu/repo -e 'using Pkg; Pkg.test()'` on Julia 1.10.11 — 1,975/1,975 passed.
- Mutation checks: disconnecting constraint scale from slack inference produced the expected compiler-test failure; removing the only zero-result guard still passed 1,975/1,975, which demonstrates the inline coverage gap.
- Constructor probe: a finite `big"1e400"` is accepted but stored as `Inf`, reproducing the other inline finding.
- GitHub/local changed-file sets match; `main` is still the branch-point commit `4fb1532`, so the tested PR head is also the current merge result.
- Current-head GitHub checks are all successful: four Julia CI jobs, docs build, labeler, and `documenter/deploy`. The docs warning set matches current `main`; no new missing-docstring warning was introduced.
- The cross-repository `Closes JuliaQUBO/QUBO.jl#67` scope and stacked #235 coordination remain appropriate.

## Summary

- Blocking: 2 (inline: stored-factor finiteness; missing zero-result regression coverage).
- Nonblocking: 1 (refresh PR metadata).
- Questions: none.
- Merge-readiness: not ready until the two blocking threads are addressed. This account is the PR author, so GitHub cannot accept a formal request-changes review. `main` requires zero approving reviews, but conversation resolution is enabled, so these inline threads are the remaining review gate.

I would not merge this until the blocking issues above are addressed.

bernalde:
Addressed all three findings from the [re-review of `403fd55`](https://github.com/JuliaQUBO/ToQUBO.jl/pull/234#pullrequestreview-4840405428) in c0b6510:

- **Blocking — stored-factor finiteness** (inline): `MultiplicativeUpdate` now converts to `Float64` first and validates the *converted* value, so `big"1e400"` raises `ArgumentError` instead of being stored as `Inf`. Regression cases cover both the overflowing `BigFloat` and a normal one (`big"2.5"` → `2.5`).
- **Blocking — zero-result stop condition** (inline): added a counting probe sampler (`RefinementProbeSampler`) plus two tests. `test_refinement_zero_result_stop` (empty from the first solve) asserts `PenaltyUpdateCount() == 0`, exactly **one** sampler call, and an untouched hint; `test_refinement_zero_result_mid_loop_stop` (violating sample, then empty) asserts exactly **one** update, **two** sampler calls, and the once-escalated hint — pinning the guard *inside* the loop, not only at entry.
- `refresh-pr-metadata` (nonblocking): rather than re-date the brittle claims, I removed them. The Testing section now describes the coverage without a count, and Branch Hygiene carries only the durable note that #235 is stacked here and merges after this PR. Re-reading live state also corrected the review's premise: the docs PR #236 it mentioned has since merged, so #235 is now the only other open PR — exactly the staleness that motivated dropping the inventory.

Tests: full local suite (`Pkg.test()`, Julia 1.10.11) green. Both fixes are mutation-verified — reverting the constructor to validate the pre-conversion value, and deleting the `ResultCount == 0` guard, each redden only their new tests (evidence in the reply threads).

**Base sync (c9ae30f):** the PR went `BEHIND` while this pass ran, so `main` (through #236) was merged in append-only and the full suite re-run green on the merged tree before pushing. The same merge was propagated to the stacked #235 so it does not inherit a stale base.

**One CI note:** the first `documenter/deploy` on the merged head failed with `! [rejected] HEAD -> gh-pages (fetch first)` — the docs *built* fine and only the preview push lost a race against #235's simultaneous preview deploy (I had pushed both branches seconds apart). Re-running that job succeeded; every current-head check is now green.

No findings were declined.

src/attributes/compiler.jl
bernalde: Addressed in c0b6510: the constructor now converts to `Float64` first and validates the stored value, so `MultiplicativeUpdate(big"1e400")` raises `ArgumentError` instead of storing `Inf`. Regression cases cover both the overflowing `BigFloat` and a clean one (`big"2.5"` → `2.5`).

Mutation-verified: reverting to validate the pre-conversion `Real` reddens exactly the new assertion (`test_refinement_attributes`, `refinement.jl:232`) with the other 1,984 tests passing.

test/unit/refinement.jl
bernalde: Addressed in c0b6510 with a counting probe sampler (`RefinementProbeSampler`) and both variants you suggested:

- `test_refinement_zero_result_stop` — empty from the first solve: asserts `PenaltyUpdateCount() == 0`, exactly **one** sampler call (no re-solve), and an untouched hint.
- `test_refinement_zero_result_mid_loop_stop` — violating sample, then empty: asserts exactly **one** update, **two** sampler calls, and the once-escalated hint (`-0.1` → `-1.0`), which pins the guard *inside* the loop rather than only at entry.

Mutation-verified with your exact experiment: deleting the `ResultCount == 0` early return in `_best_sample_violations` now makes the zero-result test error out of `optimize!` (the `_check_result_index` no-primal-results path) and abort the refinement testset, instead of passing silently as it did before.

PR #235, opened 2026-08-02
Add subgradient (augmented-Lagrangian) penalty update strategy

### LINKED ISSUE #233: Add subgradient (augmented-Lagrangian) penalty update strategy

bernalde:
Closes #233

Part of the MQT QAO parity epic: JuliaQUBO/QUBO.jl#69 (beyond-parity extension). Design record: https://github.com/JuliaQUBO/QUBO.jl/issues/67#issuecomment-5159885555

**Stacked on #234** (the refinement engine); only the last two commits are this PR's diff.

## Summary

Adds the method-of-multipliers penalty-update strategy on top of #234's engine:

- **`Attributes.AugmentedLagrangianPenalty(multiplier, rho)`** — composite `λ·ℓ + ρ·ℓ²` penalty over the constraint residual. Equality support (signed λ) plus slack-free inequalities with the linear term oriented against the infeasible direction (documented honestly as *iterated unbalanced penalization* — the one-sided Rockafellar term is not QUBO-expressible without slacks). Reuses `_unbalanced_penalty`'s emission pattern; unlike `UnbalancedPenalty`, no `ConstraintEncodingPenaltyHint` is required: `penalties!` applies exactly the sense sign σ and the coefficients live in the method.
- **`Attributes.SubgradientUpdate(; step, patience, escalation_factor)`** — per iteration, every AL constraint's multiplier updates from the best sample's **signed** residual (the #138 machinery: EqualTo `value−target`, inequalities positive-when-violated): equalities `λ ← λ + η·r`, inequalities `λ ← max(0, λ + η·r)`; η defaults to the constraint's own ρ (classical MoM). Multipliers can decrease and equality duals go negative — both tested.
- **Safeguard, redesigned for sampled problems**: the classical escalate-when-`‖r_k‖ > γ‖r_{k−1}‖` ratio test (Bertsekas γ ≈ 0.25, BCL) misfires on binary models — violation norms are quantized and sit unchanged until a constraint flips feasible, so the first smoke run escalated ρ to 10 when the multiplier alone converges. Replaced with a **patience** counter (escalate violated constraints' ρ ×`escalation_factor` only after `patience` consecutive non-improving iterations, then reset), the discrete analogue of escalate-only-when-stalled. The smoke case now shows the textbook behavior: λ → 1.2 with ρ pinned at 1.0.

## Why this matters (from the recorded study)

Multiplicative escalation needs ρ→∞ and exceeds sampler coefficient dynamic range within a few ×10 steps (Ohzeki, Sci. Rep. 10:3126, 2020, states this failure mode explicitly). Multiplier updates converge at **bounded** ρ (Powell's shifted-penalty view; Bertsekas 1982, linear rate at fixed ρ); Yonaga–Miyama–Ohzeki (arXiv:2012.06119) run exactly this iteration on a D-Wave with ρ = 0.1 fixed. This PR's headline test asserts that property: feasibility reached with `method.rho == 1.0` untouched.

## Acceptance criteria (from #233)

- [x] Under-penalized equality model reaches feasibility via multiplier updates with ρ fixed (`test_refinement_subgradient_equality`, both violation directions incl. negative duals, exact λ traces 2.25 / −1.2)
- [x] Inequality convergence with one-sided updates; multipliers floor at zero (`test_refinement_subgradient_inequality` λ-trace 0→1.2; white-box decrease/floor test)
- [x] MAX-sense model converges with correct signs (inequality + equality-from-above tests are MAX; applied coefficient σ = ±1 asserted both senses)
- [x] ρ escalation only on stalled residuals (`test_refinement_subgradient_stall_escalation`: tiny step → exactly one ×10 escalation after `patience` = 3)
- [x] Multipliers and iteration count queryable (`ConstraintEncodingMethod` readback + `PenaltyUpdateCount`); docs with references

## Testing

- `julia --project -e 'using Pkg; Pkg.test()'` (Julia 1.10.11): full suite; seven new subgradient/AL test functions with hand-derived iteration counts and multiplier values.

## Branch Hygiene

- Base: `feat/issue-67-penalty-refinement` (PR #234) — merge #234 first; this PR then retargets/merges cleanly.
- Sibling Dependabot PRs are workflow-only; no overlap.


bernalde (commented):
Maintainer review of head `15bd3ff` (author account — COMMENT review; cannot serve as formal approval). Stacked on #234; this review covers the two commits beyond that base (`9aef6fc`, plus the forward-merge of #234's address fix).

## Scope and design

Implements ToQUBO#233 per the recorded design, with one deliberate deviation the PR documents and justifies: the classical stall test (`‖r_k‖ > γ‖r_{k−1}‖`, Bertsekas/BCL) was replaced by a **patience** counter after the first smoke run demonstrated the ratio test misfires on quantized binary violations (ρ escalated to 10 in a case where the multiplier alone converges — the PR shows the corrected λ → 1.2, ρ = 1.0 trace). This is the right call for sampled problems and the docstring records the rationale. Also endorsed:

- `AugmentedLagrangianPenalty` reuses the unbalanced emission with equality support and signed multipliers; the no-hint contract (`penalties!` applies exactly σ) removes `UnbalancedPenalty`'s biggest usability trap for this use case.
- One-sided inequality updates floor at zero (white-box tested), equality duals go negative (tested from both violation directions), and the bounded-ρ headline property is asserted (`method.rho == 1.0` untouched through convergence).
- Interval/vector sets are skipped explicitly with the docstring naming the boundary.

## Findings

**Nonblocking: AL constraints are invisible in `PenaltyPolicyMetadata`** 
The `penalties!` AL branch `continue`s before any recording, so a model mixing AL and inferred constraints produces metadata whose constraint list omits the AL ones entirely — a reader auditing the reformulation cannot tell they were method-managed rather than skipped. Record a minimal entry (e.g. `kind/id/penalty = σ/selected_policy = "AugmentedLagrangianPenalty"`) in `context["inferred_penalties"]["constraints"]` so the inventory stays complete.

**Nonblocking: budget burn when violations are outside the strategy's reach** 
`SubgradientUpdate` returns `updated = true` whenever any AL multiplier moved — including λ-decreases on satisfied constraints — so a model whose *violated* constraints are all non-AL (e.g. default quadratic) can loop the full budget recompiling without any prospect of fixing those violations. Add an early break when no violated constraint is AL-managed (mirroring the multiplicative strategy's no-updatable-knob guard), or document why continuing is preferred; the trade-off is that satisfied-constraint dual decreases are legitimate but cannot resolve the violations that keep the loop alive.

## Verification performed

- Full local suite green at exactly this tree (Julia 1.10.11): seven new subgradient/AL tests with hand-derived traces (λ 0→1.2 at fixed ρ; equality duals +2.25/−1.2; single patience-gated ×10 escalation; zero-floor).
- Two standalone smoke runs during development: the bounded-ρ case and the stall case, both matching derivations after the patience redesign.
- CI green on head `15bd3ff` (full matrix + docs + labeler); docs build clean (new docstrings render in the penalty-methods and refinement sections).
- File-set and stacking verified: base branch is #234's head; only this PR's delta reviewed. `Closes #233` checked against all five acceptance criteria — satisfied at this head.

## Summary

- Blocking: none.
- Nonblocking: 2 (AL metadata entry, unreachable-violation break).
- Questions: none.
- Merge-readiness: CI green, `CLEAN` against its stacked base; merge order is #234 first. This COMMENT review cannot formally approve.

bernalde:
Addressed both nonblocking findings from the [maintainer review](https://github.com/JuliaQUBO/ToQUBO.jl/pull/235#pullrequestreview-4839557385), landing across f6ec612 + 2435299:

- `al-metadata-entry` — `penalties!` now records each AL-managed constraint in `PenaltyPolicyMetadata`'s constraint inventory (`selected_policy = "AugmentedLagrangianPenalty"`, applied σ as the penalty); asserted in `test_refinement_al_compiles_without_hint` for both senses.
- `unreachable-violation-break` — the loop now breaks when no *violated* constraint is AL-managed, via a strategy-dispatched `_reachable_violation` predicate in the **engine**. The first attempt (f6ec612) placed the guard inside `_apply_penalty_update!`, which broke the white-box satisfied-decrease contract (4 test failures) and was pushed prematurely — 2435299 relocates the guard to the engine where loop-progress semantics belong, restoring the updater's pure contract. New engine-level test `test_refinement_subgradient_unreachable_violation_break`: a violated quadratic constraint plus a satisfied AL constraint yields count 0 with the AL multiplier untouched.

Tests: full local suite (`Pkg.test()`, Julia 1.10.11) green at 2435299, zero failures (f6ec612 was transiently red on the remote; superseded). CI is running on the new head.

No findings were declined.

src/attributes/compiler.jl
```diff
@@ -309,6 +352,54 @@ struct MultiplicativeUpdate <: PenaltyUpdate
...
+        escalation_factor::Real = 10.0,
+    )
+        isnothing(step) || (isfinite(step) && step > 0) ||
+            throw(ArgumentError("step must be finite and positive"))
+        patience >= 1 || throw(ArgumentError("patience must be at least one"))
+        isfinite(escalation_factor) && escalation_factor > 1 ||
+            throw(ArgumentError("escalation factor must be finite and greater than one"))
+
+        return new(
+            isnothing(step) ? nothing : Float64(step),
```
bernalde: **Blocking: validate the values that `SubgradientUpdate` actually stores** 

The checks above run on the arbitrary-precision `Real` inputs, but this line and the escalation-factor line then narrow them to `Float64`. At exact head `4401366`, `SubgradientUpdate(step = big"1e400")` stores `Inf`, `step = big"1e-400"` stores `0.0`, and `escalation_factor = big"1e400"` stores `Inf`; those values passed the documented finite/positive contracts before conversion and then make refinement error mid-loop or silently stop updating. The newly merged base fixes the identical `MultiplicativeUpdate` defect by converting first and validating the stored value. Please apply that pattern to both optional `step` and `escalation_factor`, with overflow and underflow regressions.

docs/src/manual/4-settings.md
```diff
@@ -304,17 +304,23 @@ Each iteration fully recompiles the model;
 only, so total refinement overhead scales with the iteration count.
 
 Escalation grows the QUBO's coefficient range, which physical samplers resolve
-with limited precision; a bounded-coefficient subgradient (augmented
-Lagrangian) strategy is tracked in
-[ToQUBO#233](https://github.com/JuliaQUBO/ToQUBO.jl/issues/233). Violations
-driven by variable encodings rather than constraint penalties have no
-per-constraint coefficient and are not escalated.
+with limited precision. For bounded coefficients, configure constraints with
+[`ToQUBO.Attributes.AugmentedLagrangianPenalty`](@ref) and select
```
bernalde: **Nonblocking: remove the PR-introduced Documenter size warning** 

The exact-head Documentation job is green but emits `Generated HTML over size_threshold_warn limit: manual/4-settings.md` at the 100 KiB threshold; the completed current-base job at `c9ae30f` does not. This new section tips an already large settings page over the configured warning limit. Please split the refinement material into a focused page (and wire it into navigation) or otherwise bring this page back under the threshold, then confirm the warning disappears from the docs log.

bernalde (commented):
Maintainer re-review of head `4401366597823236f501d5e1c09e40d830fc904a` (author account — COMMENT review; cannot serve as formal approval or request-changes state). This pass re-reviewed every commit since the prior `15bd3ff` sentinel, the full six-file PR diff, and the tree after the latest stacked-base merge.

## Prior findings

Both findings from the earlier review are addressed:

- AL-managed constraints are now included in `PenaltyPolicyMetadata`. I removed only that recording in a disposable checkout; `test_refinement_al_compiles_without_hint` then errored on the empty inventory, confirming the regression test guards the fix.
- The engine now stops when none of the violated constraints is reachable by `SubgradientUpdate`. I removed only the `_reachable_violation` guard; the targeted test then observed one unwanted update and a multiplier change from 1.0 to 0.0, confirming the engine-level test guards both loop behavior and state preservation.

## Findings

The constructor correctness issue and the Documenter warning regression are attached inline.

**Nonblocking: refresh the stale scope and testing claims in the PR body** 

The body still says that only the last two commits form this PR and that there are seven new subgradient/AL test functions. The live history now includes the feature commit, two address commits, and two stacked-base merge commits, while `test/unit/refinement.jl` registers eight new functions including the unreachable-violation regression. Please update those statements so the review basis and test inventory match the current head.

## Verification

- Full local suite at exact head `4401366`, Julia 1.10.11: `2029 / 2029` passed (`julia --project -e 'using Pkg; Pkg.test()'`).
- Targeted constructor probe: finite `BigFloat` inputs currently produce stored values `(step = Inf)`, `(step = 0.0)`, and `(escalation_factor = Inf)` after `Float64` conversion.
- Current-head CI is green across Julia 1.10/latest on Ubuntu/Windows, Documentation, labeler, and `documenter/deploy`.
- Documentation-log comparison: the exact-head job newly emits `Generated HTML over size_threshold_warn limit: manual/4-settings.md`; the current stacked-base job at `c9ae30f` does not.
- GitHub and local changed-file sets match (six files); `git diff --check` is clean; the head contains the current stacked base exactly (`merge-base = c9ae30f`).
- Linked issue #233 and its QUBO#67 design record were checked against all acceptance criteria. No unrelated open PR overlaps this six-file delta; #234 is the expected stacked base.

## Summary

- Blocking: 1 (validate the converted values stored by `SubgradientUpdate`).
- Nonblocking: 2 (remove the new docs-size warning; refresh stale PR-body scope/testing metadata).
- Questions: 0.
- Prior findings: 2 addressed, 0 declined.
- Merge readiness: checks and the stacked merge are green, but the constructor accepts invalid stored state. The feature base branch is unprotected, so this author COMMENT is advisory rather than an enforced approval gate.

I would not merge this until the blocking issues above are addressed.

bernalde:
Addressed both findings from the [re-review of `4401366`](https://github.com/JuliaQUBO/ToQUBO.jl/pull/235#pullrequestreview-4840530893) in 6c84370, on top of a base sync (573f247).

- **Blocking — `subgradient-converted-values`**: reproduced all three cases at `4401366` before fixing (`step = big"1e400"` → `Inf`, `step = big"1e-400"` → `0.0`, `escalation_factor = big"1e400"` → `Inf`). `SubgradientUpdate` now converts `step` and `escalation_factor` to `Float64` first and validates the stored values, matching the pattern the base uses for `MultiplicativeUpdate`. Regressions cover overflow **and** underflow for `step`, overflow for `escalation_factor`, plus clean-conversion acceptance (`big"0.25"` → `0.25`, `big"2.5"` → `2.5`).
- **Nonblocking — `settings-page-size-warning`**: confirmed the warning was real and PR-introduced (present in the `4401366` docs job, absent at base `c9ae30f`), then split the refinement material into a dedicated `manual/5-refinement.md` and wired it into the `make.jl` navigation. Verified locally by building the docs: the build exits 0 with **no `size_threshold` line**, and `manual/4-settings/index.html` is now **91.6 KiB** (under the 100 KiB warn limit) with the new page at 18.4 KiB. Caveat worth knowing: 4-settings now sits only ~8 KiB under the threshold, so the next sizeable addition there will re-trip it.

**`AugmentedLagrangianPenalty` — scope note and follow-up fix (4e836bb).** The constructor-finiteness pattern deliberately was *not* applied to this type, because it does not have that defect: it is parametric (`{T}` via `promote`), so it validates exactly the values it stores, and narrowing it to `Float64` would break legitimate `Optimizer{BigFloat}` models.

Probing it did surface a real defect, which the author has since authorized fixing here: a method parameterized on a **wider** type than the model's coefficients (a `BigFloat` method on a `Float64` model) built a malformed pseudo-Boolean function and died with a `MethodError` in `zero(...)`. The same failure reproduced with the pre-existing `UnbalancedPenalty(big"1.0", big"0.5")`, so both are fixed together: a shared `_method_coefficient` helper now converts each method coefficient to the model's numeric type at all four emission sites (augmented-Lagrangian and unbalanced, `LessThan` and `GreaterThan`) and validates the *converted* value — the same "validate what you store" principle as the constructor findings, applied at the second boundary that violated it.

Verified: `AugmentedLagrangianPenalty(big"0.5", big"1.0")` now produces exactly the same objective as its `Float64` equivalent; an overflowing multiplier and an underflowing `rho` each raise a descriptive `CompilationError` instead of silently becoming `Inf`/`0.0`; and the previously-broken `UnbalancedPenalty{BigFloat}` compiles to the same coefficients as the defaults. Regressions live in `test/unit/refinement.jl` and `test/unit/unbalanced_penalty.jl`, and the docstring now states the conversion contract.

Tests: full local suite (`Pkg.test()`, Julia 1.10.11) green at 4e836bb; local docs build green (no `size_threshold` line, and CI's docs job at 6c84370 confirmed the same). CI on the new head is running.

No findings were declined.

src/attributes/compiler.jl
bernalde: Addressed in 6c84370. I reproduced all three cases at `4401366` first (`step = big"1e400"` → `Inf`, `step = big"1e-400"` → `0.0`, `escalation_factor = big"1e400"` → `Inf`), then applied the base's `MultiplicativeUpdate` pattern: convert `step`/`escalation_factor` to `Float64` up front and validate the stored values. Regressions cover overflow and underflow for `step`, overflow for `escalation_factor`, and clean-conversion acceptance (`big"0.25"` → `0.25`).

Not extended to `AugmentedLagrangianPenalty`: it is parametric (`{T}` via `promote`), so it already validates exactly what it stores — narrowing it to `Float64` would break legitimate `Optimizer{BigFloat}` models. Details and a separate pre-existing finding are in the [summary comment](https://github.com/JuliaQUBO/ToQUBO.jl/pull/235#issuecomment-5167724124).

docs/src/manual/4-settings.md
bernalde: Addressed in 6c84370: the refinement material now lives in a dedicated `docs/src/manual/5-refinement.md`, wired into the `make.jl` navigation as "Penalty Refinement".

Confirmed as you asked, by building the docs locally rather than inferring it: the build exits 0 with **no `size_threshold` line in the log**, and `manual/4-settings/index.html` is now **91.6 KiB** — under the 100 KiB `size_threshold_warn` — with the new page at 18.4 KiB. One caveat: 4-settings is only ~8 KiB under the limit now, so the next sizeable addition to that page will re-trip the warning.

